PR #2, opened 2019-04-05
Changing contact point for keystone to be public address.

hyperbolic2346:
Private address will not always work for cross-model relations.

Fixes [lp:1822893](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1822893)


hyperbolic2346:
Done before I made the PR. It works and is valid in both situations.

PR #5, opened 2019-04-15
Adding support for setting the virtual IP that points to the control plane

hyperbolic2346:
This allows for someone with an external load balancer to specify the IP that the workers will use to talk to the control plane.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -267,13 +267,12 @@ options:
   loadbalancer-ips:
     type: string
     description: |
-      Space seperated list of IP addresses of loadbalancers in front of control plane.
-      These can be either virtual IP addresses that are floated in front of the control
-      plane or the IP of a loadbalancer such as an F5. The workers will alternate IP
-      addresses from this list to distribute load. If you have 2 IPs and 4 workers,
-      each IP will be used by 2 workers. Note that this will only work if there is a
-      relation between kubernetes-master:kube-api-endpoint and
-      kubernetes-worker:kube-api-endpoint. This is the case if kubeapi-load-balancer
-      is not in use. If it is, use the loadbalancer-ips configuration variable on
-      the kubeapi-load-balancer charm.
+      Space separated list of IP addresses of loadbalancers in front of the control plane.
+      These can be either virtual IP addresses that have been floated in front of the control
+      plane or the IP of a loadbalancer appliance such as an F5. Workers will alternate IP
+      addresses from this list to distribute load - for example If you have 2 IPs and 4 workers,
+      each IP will be used by 2 workers. Note that this will only work if kubeapi-load-balancer
+      is not in use and there is a relation between kubernetes-master:kube-api-endpoint and
+      kubernetes-worker:kube-api-endpoint. If using the kubeapi-load-balancer, see the
+      loadbalancer-ips configuration variable on the kubeapi-load-balancer charm.
     default: ""
--- a/config.yaml
+++ b/config.yaml
@@ -264,3 +264,12 @@ options:
       "basic", and "token". If set to "auto", basic auth is used unless Keystone is 
       related to kubernetes-master, in which case token auth is used.
     default: "auto"
+  loadbalancer-ips:
+    type: string
+    description: |
+      Space seperated list of IP addresses of loadbalancers in front of control plane.
+      These can be either virtual IP addresses that are floated in front of the control
+      plane or the IP of a loadbalancer such as an F5. The workers will alternate IP
+      addresses from this list to distribute load. If you have 2 IPs and 4 workers,
+      each IP will be used by 2 workers.
+    default: ""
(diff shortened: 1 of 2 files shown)


hyperbolic2346:
Moved from https://github.com/juju-solutions/kubernetes/pull/226

Fixes https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/629

Cynerva:
> This allows for someone with an external load balancer to specify the IP that the workers will use to talk to the control plane.

Not quite. This affects kubeconfig files generated on kubernetes-master, but not on kubernetes-worker. The kubernetes-worker charm creates its own kubeconfigs using addresses it receives over the kube-api-endpoint relation, which _might_ be related directly to kubernetes-master, but is usually related to kubeapi-load-balancer instead.

hyperbolic2346:
A good point, sir. I've updated the PR to include updating the relation data. I was on the fence about doing something like this for the kube-api-loadbalancer charm, but it is possible to have a virtual IP floating on the load balancers outside of juju and wanting to use that. As a result, I am going to add a similar PR to the load balancer charm. This will only work if there isn't a load balancer. I updated the config.yaml to spell that out.

@evilnick I would love your input on the wording on the config.yaml section if you have any.

reactive/kubernetes_master.py
```diff
@@ -1008,12 +1021,23 @@ def loadbalancer_kubeconfig():
...
-        # Get the public address of the first loadbalancer so
-        # users can access the cluster.
-        address = hosts[0].get('public-address')
+        if hacluster_vip:
+            address = hacluster_vip
+        else:
+            if len(hosts) < 1:
+                # oops...hopefully we just lost our relation.
+                # skip this for now.
+                return
```
Cynerva: If the `loadbalancer.available` flag is set, we shouldn't ever encounter this case. If we do, that should be fixed in interface-public-address.

I'm OK with having a check here too just in case -- hook errors are nasty -- but please include a log message so we know it happened.

reactive/kubernetes_master.py
```diff
@@ -2310,3 +2339,15 @@ def get_dns_provider():
...
+@when('config.changed.loadbalancer-ips')
+def vip_changed():
+    # get a new cert
+    if (is_state('certificates.available') and
+            is_state('kube-api-endpoint.available')):
+        send_data()
+
+    # update workers
+    if (is_state('kube-api-endpoint.available')):
+        push_service_data()
```
Cynerva: No need for this handler. `send_data` and `push_service_data` run on every hook invocation anyway.

reactive/kubernetes_master.py
hyperbolic2346: I'm fine with taking this out of this PR and getting more information. I just saw an error here when I removed the relation and thought I would try and fix it. Want me to just toss it for now and add a bug?

reactive/kubernetes_master.py
hyperbolic2346: Wouldn't this handler prevent a "change the config and wait 5 minutes for a hook" or are hooks guaranteed to run on config change?

Cynerva: When the config changes, Juju calls the charm's `config-changed` hook, which runs reactive, which runs any handlers whose decorator conditions are met. Both `send_data` and `push_service_data` have conditions that lead them to run repeatedly, so in this case, they will run during the `config-changed` hook and use the new config value. No 5 min delay.

reactive/kubernetes_master.py
Cynerva: Yeah, that works for me :+1:

Cynerva:
LGTM

Cynerva:
Feel free to merge this. I left it unmerged in case you wanted to wait for feedback on the config description.

config.yaml
```diff
@@ -264,3 +264,16 @@ options:
       "basic", and "token". If set to "auto", basic auth is used unless Keystone is 
       related to kubernetes-master, in which case token auth is used.
     default: "auto"
+  loadbalancer-ips:
+    type: string
+    description: |
+      Space seperated list of IP addresses of loadbalancers in front of control plane.
```
tvansteenburgh: Space-separated

evilnick: I would go with

Space separated list of IP addresses of loadbalancers in front of the control plane.
These can be either virtual IP addresses that have been floated in front of the control
plane or the IP of a loadbalancer appliance such as an F5. Workers will alternate IP
addresses from this list to distribute load - for example If you have 2 IPs and 4 workers,
each IP will be used by 2 workers.  Note that this will only work if kubeapi-load-balancer
is not in use and there is a relation between kubernetes-master:kube-api-endpoint and
kubernetes-worker:kube-api-endpoint. If using the kubeapi-load-balancer, see the 
loadbalancer-ips configuration variable on the kubeapi-load-balancer charm.

tvansteenburgh:
@hyperbolic2346 We should add something about this to the High Availability section of the docs (https://www.ubuntu.com/kubernetes/docs/high-availability)

PR #7, opened 2019-04-22
Fix stucking at "Waiting for master components to start" after upgrading

xtrusia:
install_snap() removes flag kubernetes-master.components.started
but there is no set it back.

so status is stuck at "Waiting for master components to start"

##
I just removed it but if there is proper way, Please advice me.

Thanks


xtrusia:
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1825819

Cynerva:
This isn't the right fix, since it will prevent the Kubernetes services from being reconfigured properly after snap upgrades.

Please refer back to https://bugs.launchpad.net/charm-kubernetes-master/+bug/1825819. We'll help you there.

PR #10, opened 2019-04-25
Make setup_basic_auth() non-destructive.

tvansteenburgh:
Fixes [lp:1826260](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1826260)

This changes the behavior of setup_basic_auth(). Instead of overwriting
the entire /root/cdk/basic_auth.csv file and replacing it with a single
line, it will now:

1. Update a single line if the file already contains an entry for
`username`.
2. Add a single line if the file does not already contain an entry
for `username`.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -52,7 +52,7 @@ from charms.layer import vault_kv
 from charmhelpers.core import hookenv
 from charmhelpers.core import host
 from charmhelpers.core import unitdata
-from charmhelpers.core.host import service_stop
+from charmhelpers.core.host import service_stop, service_resume
 from charmhelpers.core.templating import render
 from charmhelpers.fetch import apt_install
 from charmhelpers.contrib.charmsupport import nrpe
@@ -645,6 +645,34 @@ def add_systemd_restart_always():
         copyfile(template, '{}/always-restart.conf'.format(dest_dir))
 
 
+def add_systemd_file_watcher():
+    """Setup systemd file-watcher service.
+
+    This service watches these files for changes:
+
+    /root/cdk/basic_auth.csv
+    /root/cdk/known_tokens.csv
+    /root/cdk/serviceaccount.key
+
+    If a file is changed, the service uses juju-run to invoke a script in a
+    hook context on this unit. If this unit is the leader, the script will
+    call leader-set to distribute the contents of these files to the
+    non-leaders so they can sync their local copies to match.
+
+    """
+    render('cdk.master.leader.file-watcher.sh',
+        '/usr/local/sbin/cdk.master.leader.file-watcher.sh',
+        {}, perms=0o777)
+    render('cdk.master.leader.file-watcher.service',
+        '/etc/systemd/system/cdk.master.leader.file-watcher.service',
+        {'unit': hookenv.local_unit()}, perms=0o644)
+    render('cdk.master.leader.file-watcher.path',
+        '/etc/systemd/system/cdk.master.leader.file-watcher.path',
+        {}, perms=0o644)
+    service_resume('cdk.master.leader.file-watcher.service')
+    service_resume('cdk.master.leader.file-watcher.path')
+
+
 @when('etcd.available', 'tls_client.certs.saved',
       'authentication.setup',
       'leadership.set.auto_storage_backend',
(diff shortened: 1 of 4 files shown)


reactive/kubernetes_master.py
```diff
@@ -1720,19 +1721,35 @@ def configure_scheduler():
...
-        else:
-            stream.write('{0},{1},{2}'.format(password, username, uid))
+
+    newline = '{0},{1},{2}'.format(password, username, uid)
+    if groups:
+        newline += ',"{0}"'.format(groups)
+
+    # If we have an existing line for this username, update it, leaving
+    # lines for other users unchanged.
+    with fileinput.input(htaccess, inplace=True) as f:
```
johnsca: I hadn't heard of this before.  Nice.

reactive/kubernetes_master.py
```diff
@@ -1720,19 +1721,35 @@ def configure_scheduler():
...
+            if (line_user, line_uid) == (username, uid):
+                print(newline)
+                user_written = True
+            else:
+                print(line.strip())
+
+    # If we didn't find an existing line for this username, add one.
+    if not user_written:
+        with open(htaccess, 'a') as f:
+            f.write(newline)
```
johnsca: Seems like it would have been a bit cleaner to nest this under the context manager and just use `print()`.

reactive/kubernetes_master.py
```diff
@@ -1720,19 +1721,35 @@ def configure_scheduler():
...
-            stream.write('{0},{1},{2}'.format(password, username, uid))
+
+    newline = '{0},{1},{2}'.format(password, username, uid)
+    if groups:
+        newline += ',"{0}"'.format(groups)
+
+    # If we have an existing line for this username, update it, leaving
+    # lines for other users unchanged.
+    with fileinput.input(htaccess, inplace=True) as f:
+        user_written = False
```
johnsca: Oh, and the name of this variable confused me for a second.  I might suggest `wrote_user_line` instead.

reactive/kubernetes_master.py
tvansteenburgh: Ah, yeah, that's what I tried first, but it doesn't work. You can't append to the file. I found it weird and unexpected, but apparently it's by design, since the docs say: "With mode you can specify which file mode will be passed to open(). It must be one of 'r', 'rU', 'U' and 'rb'."

reactive/kubernetes_master.py
tvansteenburgh: Thanks, updated.

reactive/kubernetes_master.py
tvansteenburgh: Actually, that doesn't really explain it. Clearly the file is being written. But regardless, I stepped through it in the debugger, and although the print call was made, the extra line was not written to the file. It seemed that with fileinput, you can write an existing line, or change an existing line, but you can't add a new line.

tvansteenburgh: Yeah, so I read the source for fileinput. When the loop through the lines finishes, the output file is closed, so you can't write any additional lines.

johnsca: Very strange.  It seems that fileinput with `inplace=True` only does the `stdout` redirect when actually iterating over the lines, rather than during the entire context manager block:

```python
In [22]: filepath.write_text('one\ntwo\nthree')                                                                                                                                                                                  
Out[22]: 13

In [23]: print(filepath.read_text())                                                                                                                                                                                             
one
two
three

In [24]: with fileinput.input(str(filepath), inplace=True) as f: 
    ...:     print('new - start') 
    ...:     for l in f: 
    ...:         print(l.strip()) 
    ...:         print(l.strip()) 
    ...:     print('new - end') 
    ...:                                                                                                                                                                                                                         
new - start
new - end

In [25]: print(filepath.read_text())                                                                                                                                                                                             
one
one
two
two
three
three


In [26]: 
```

I guess that makes sense from the perspective of it being able to process multiple files with a single loop.

johnsca:
A bit more bike-shedding:

```python
import csv

def setup_basic_auth(password=None, username='admin', uid='admin',
                     groups=None):
    '''Add or update an entry in /root/cdk/basic_auth.csv.'''
    htaccess = Path('/root/cdk/basic_auth.csv')
    htaccess.parent.mkdir(parents=True, exist_ok=True)

    if not password:
        password = token_generator()

    new_row = [password, username, uid] + ([groups] if groups else [])

    with htaccess.open('r') as f:
        rows = list(csv.reader(f))

    for row in rows:
        if row[1] == username or row[2] == uid:
            # update existing entry based on username or uid
            row[:] = new_row
            break
    else:
        # append new entry
        rows.append(new_row)

    with htaccess.open('w') as f:
        csv.writer(f).writerows(rows)
```

tvansteenburgh:
This is ready for another review.

PR #12, opened 2019-04-25
Fix kubelet-preferred-address-types set incorrectly

Cynerva:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1826397

@johnsca should be able to vouch for this as a fix for `kubectl logs` failing on openstack.

Some additional evidence: [kubelet-preferred-address-types in the Kubernetes code](https://github.com/kubernetes/kubernetes/blob/b7394102d6ef778017f2ca4046abbaa23b88c290/cmd/kube-apiserver/app/options/options.go#L190) is defined using StringSliceVar. Kubernetes uses [Cobra](https://github.com/spf13/cobra), which uses [pflag](https://github.com/spf13/pflag), which documents StringSliceVar [here](https://godoc.org/github.com/spf13/pflag#StringSliceVar). No brackets.


ktsakalozos:
LGTM +1

PR #15, opened 2019-04-29
Switch to external OpenStack cloud provider

johnsca:
The in-tree cloud providers are being transitioned to external providers using the cloud-controller-manager plugin model.  The in-tree OpenStack provider has already been deprecated, so this switches the charms to use the external provider and the Cinder CSI plugin.

Depends on: https://github.com/charmed-kubernetes/cdk-addons/pull/99
Depends on: https://github.com/charmed-kubernetes/layer-kubernetes-common/pull/2
Depends on: https://github.com/charmed-kubernetes/cdk-addons/pull/108


johnsca:
@hyperbolic2346 Can you take another look at this.

PR #24, opened 2019-05-28
Configure master with an image-registry

kwmonroe:
Per https://bugs.launchpad.net/charm-kubernetes-master/+bug/1828853, we have a few commits to cdkaddons and interface-kube-control that will facilitate a better way to configure a container registry for cdk.

Pick up where @hyperbolic2346 started to bring k8s-master in line with that vision.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -50,8 +50,8 @@ options:
       Specify the docker registry to use when applying addons.
 
       DEPRECATED in 1.15: Use the broader 'image-registry' config option instead. If both
-      options are set, addons-registry will take precedence during cdk-addons snap
-      configuration until v1.14 reaches EOL.
+      options are set, 'addons-registry' will be used to configure the cdk-addons snap until
+      v1.17 is released. After that, the 'addons-registry' option will have no effect.
   image-registry:
     type: string
     default: "image-registry.canonical.com:5000/cdk"
--- a/config.yaml
+++ b/config.yaml
@@ -43,10 +43,6 @@ options:
       Audit webhook config passed to kube-apiserver via --audit-webhook-config-file.
       For more info, please refer to the upstream documentation at
       https://kubernetes.io/docs/tasks/debug-application-cluster/audit/
-  addons-registry:
-    type: string
-    default: ""
-    description: Specify the docker registry to use when applying addons
   enable-dashboard-addons:
     type: boolean
     default: True
@@ -264,3 +260,10 @@ options:
       "basic", and "token". If set to "auto", basic auth is used unless Keystone is 
       related to kubernetes-master, in which case token auth is used.
     default: "auto"
+  image-registry:
+    type: string
+    description: |
+      Container image registry to use for CDK. This includes addons like the Kubernetes dashboard,
+      metrics server, ingress, and dns along with impacting the source location for the pause
+      container and default backend image.
+    default: "http://image-registry.canonical.com:5000/cdk"
--- a/config.yaml
+++ b/config.yaml
@@ -47,15 +47,18 @@ options:
     type: string
     default: ""
     description: |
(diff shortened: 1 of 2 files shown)


config.yaml
```diff
@@ -46,7 +46,19 @@ options:
   addons-registry:
     type: string
     default: ""
-    description: Specify the docker registry to use when applying addons
+    description: |
+      Specify the docker registry to use when applying addons.
+
+      DEPRECATED in 1.15: Use the broader 'image-registry' config option instead. If both
+      options are set, addons-registry will take precedence during cdk-addons snap
+      configuration until v1.14 reaches EOL.
```
hyperbolic2346: Should we do this or make addons-registry takes precedent in < 1.15? Seems it would be an easy check and also not rely on us later remembering to change it.

reactive/kubernetes_master.py
```diff
@@ -2402,3 +2408,18 @@ def get_dns_provider():
...
+@when('kube-control.connected')
+@when_not('kubernetes-master.sent-registry')
+def send_registry_location():
+    registry_location = hookenv.config('image-registry')
+    kube_control = endpoint_from_flag('kube-control.connected')
+    kube_control.set_registry_location(registry_location)
+    set_flag('kubernetes-master.sent-registry')
+
+
+@when('kube-control.connected')
```
hyperbolic2346: Do we want to rely on kube-control being connected here? I can contrive a situation where this isn't true and you change the config and then it isn't properly reflected. It's reaching, but also an easy change.

config.yaml
kwmonroe: @hyperbolic2346 i kept `addons-registry` taking precedent in case people already had it set in their deployment.  If it's empty (the default), `image-registry` will take over -- if it's not empty, we'll only use it for addons, with the rest of the images (pause, default backend, etc) still expected to come from `image-registry`.

reactive/kubernetes_master.py
```diff
@@ -2402,3 +2408,18 @@ def get_dns_provider():
 
     leader_set(auto_dns_provider=dns_provider)
     return dns_provider
+
+
+@when('kube-control.connected')
```
kwmonroe: @hyperbolic2346 one thing i hadn't considered until now... Should we also gate `send_registry_location` with an `is_leader` flag?

I'm not sure the ramifications of multiple masters sending registry locations to the workers, but I imagine it'll at least mean the workers/masters exchange unnecessary chatter.  Seems like we'd be better served if only the lead master send the registry config.

config.yaml
hyperbolic2346: No, I get that and like it. My question is in the "this takes precedent until we change it after 1.14 is EOL" vs "this takes precedent if the api server version is < 1.15". It is a subtle difference of hard-coded to always use addons-registry first or prefer different things based on api server version. Maybe it should be if api server version is < 1.17 to allow upgrade time? I certainly don't want to yank the rug out from under people.

reactive/kubernetes_master.py
hyperbolic2346: @kwmonroe unnecessary chatter is not good, but would we need to be concerned with units that used to be leaders still having data set on the relation?

kwmonroe: Ah, right.. without relation-setting (or un-setting, in this case), a leader could be configured, send data, then be reconfigured, lose leadership, then have workers asking about that data from a non-leader -- without an anti handler, workers would continue getting invalid config. At least with all masters talking about the same config, all workers will get it right.

config.yaml
kwmonroe: ahhh, roger that @hyperbolic2346.  I've updated the config description and added a 1.17 check to address this.

kwmonroe:
> Test coverage?

https://github.com/charmed-kubernetes/jenkins/issues/419

PR #29, opened 2019-06-17
lp 1833089: only call leader_set when is_leader

kwmonroe:
There's been a change in 2.6.4 that turned a warning into an error when followers attempted to call leader-set.  This led to:

https://bugs.launchpad.net/charm-kubernetes-master/+bug/1833089

`get_dns_provider` will be called by a follower during `set_final_status`. Ensure we only call `leader_set` if we're the leader. It's safe for followers to return from `get_dns_provider` without calling `leader_set` because:

- followers weren't actually setting leader data before now
- if no config nor leader_get data is available, the same logic is used for all units to set `dns_provider = core-dns`

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -2406,7 +2406,10 @@ def get_dns_provider():
             else:
                 dns_provider = 'kube-dns'
 
-    leader_set(auto_dns_provider=dns_provider)
+    # LP: 1833089. Followers end up here when setting final status; ensure only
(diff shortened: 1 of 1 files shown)


kwmonroe:
Local test with `juju 2.6.4+2.6-6c4f52c     8194   2.6/candidate  canonical✓  classic` looks good.  Merging to get this through CI.

kwmonroe:
released in cs:~containers/kubernetes-master-694

PR #32, opened 2019-06-21
Fix followers calling leader-set when using keystone

Cynerva:
`configure_cdk_addons` is only supposed to be called by leaders, but `regenerate_cdk_addons` was giving followers a back door into it.

The `configure_cdk_addons` handler runs on every hook anyway. Fix this by removing `regenerate_cdk_addons`.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -925,13 +925,6 @@ def kick_api_server():
     clear_flag('tls_client.certs.changed')
 
 
-@when_any('config.changed.keystone-policy',
-          'config.changed.keystone-ssl-ca')
-@when('keystone-credentials.available.auth')
-def regenerate_cdk_addons():
-    configure_cdk_addons()
-
-
 @when_any('kubernetes-master.components.started',
           'kubernetes-master.ceph.configured',
           'keystone-credentials.available.auth')


hyperbolic2346:
won't this delay the configuring of cdk-addons for, at most, 5 minutes?

Cynerva:
It shouldn't. When the keystone-policy or keystone-ssl-ca configs change, Juju will immediately run the config-changed hook on all of the kubernetes-master units. That will run all the reactive handlers whose flag conditions are met.

If you look at the flag conditions for the `configure_cdk_addons` handler:

https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/9ccd758a571034b6a66e6d74e30ff1038559843e/reactive/kubernetes_master.py#L928-L932

You'll see that
1. On the leader, `leadership.is_leader` is always set.
2. In the case we're concerned about, `keystone-credentials.available.auth` is set. (otherwise, `regenerate_cdk_addons` would not have run anyway)

That's enough to meet the conditions. So, it'll run on the config-changed hook, using the new config values, without having to wait for a later update-status hook or whatever.

hyperbolic2346:
@Cynerva thanks, I always forget that a config change will cause a hooks run. I get caught up in the `@when` flags and incorrectly think that only things that call out that config will run.

PR #34, opened 2019-06-27
Handle updates to certs or CA info

johnsca:
If the certificates or CA info is updated (re-issued, transferred to a different CA, etc), the charm needs to handle that by updating the files on disk (already done) and restarting the services (added here).

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -745,6 +745,11 @@ def etcd_data_change(etcd):
     if data_changed('etcd-connect', connection_string):
         remove_state('kubernetes-master.components.started')
 
+    # If the cert info changes, remove the started state to trigger
+    # handling of the master components
+    if data_changed('etcd-certs', etcd.get_client_credentials()):
+        remove_state('kubernetes-master.components.started')
+
     # We are the leader and the auto_storage_backend is not set meaning
     # this is the first time we connect to etcd.
     auto_storage_backend = leader_get('auto_storage_backend')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -2466,7 +2466,7 @@ def restart_addons_for_ca():
         service_accounts = []
         for namespace, name in service_account_names:
             output = kubectl(
-                'kubectl', 'get', 'ServiceAccount', name,
+                'get', 'ServiceAccount', name,
                 '-o', 'json',
                 '-n', namespace
             ).decode('UTF-8')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -732,7 +732,7 @@ def ca_written():
     remove_state('kubernetes-master.components.started')
     if is_state('leadership.is_leader'):
         if leader_get('kubernetes-master-addons-ca-in-use'):
-            leader_set({'kubernetes-master-addons-restart-needed': True})
+            leader_set({'kubernetes-master-addons-restart-for-ca': True})
     remove_state('tls_client.ca.written')
 
 
@@ -2440,23 +2440,87 @@ def send_new_registry_location():
 
 
 @when('leadership.is_leader',
-      'leadership.set.kubernetes-master-addons-restart-needed',
+      'leadership.set.kubernetes-master-addons-restart-for-ca',
       'kubernetes-master.components.started')
-def restart_addons():
+def restart_addons_for_ca():
     try:
-        cmd = [
-            'kubectl', 'get',
-            '-o', 'json',
-            '--all-namespaces',
-            '-l', 'cdk-addons=true',
-            'daemonset,deployment,statefulset'
+        # Get deployments/daemonsets/statefulsets
+        labels = ['cdk-addons=true', 'cdk-ingress=true']
+        deployments = []
+        for label in labels:
+            cmd = [
+                'kubectl', 'get',
+                '-o', 'json',
+                '--all-namespaces',
+                '-l', label,
+                'daemonset,deployment,statefulset'
+            ]
+            output = check_output(cmd).decode('UTF-8')
+            deployments += json.loads(output)['items']
+
+        # Get ServiceAccounts
+        service_account_names = set(
+            (
+                deployment['metadata']['namespace'],
+                deployment['spec']['template']['spec'].get(
+                    'serviceAccountName', 'default'
+                )
+            )
+            for deployment in deployments
+        )
+        service_accounts = []
+        for namespace, name in service_account_names:
+            cmd = [
+                'kubectl', 'get',
+                '-o', 'json',
+                '-n', namespace,
+                'ServiceAccount', name
+            ]
+            output = check_output(cmd).decode('UTF-8')
+            service_account = json.loads(output)
+            service_accounts.append(service_account)
+
+        # Get ServiceAccount secrets
+        secret_names = set()
+        for service_account in service_accounts:
+            namespace = service_account['metadata']['namespace']
+            for secret in service_account['secrets']:
+                secret_names.add((namespace, secret['name']))
+        secrets = []
+        for namespace, name in secret_names:
+            cmd = [
+                'kubectl', 'get',
+                '-o', 'json',
+                '-n', namespace,
+                'Secret', name
+            ]
+            output = check_output(cmd).decode('UTF-8')
+            secret = json.loads(output)
+            secrets.append(secret)
+
+        # Check secrets have updated CA
+        with open(ca_crt_path, 'rb') as f:
+            ca = f.read()
+        encoded_ca = base64.b64encode(ca).decode('UTF-8')
+        mismatched_secrets = [
+            secret for secret in secrets
+            if secret['data']['ca.crt'] != encoded_ca
         ]
-        output = check_output(cmd).decode('UTF-8')
-        items = json.loads(output)['items']
-        for item in items:
-            kind = item['kind']
-            namespace = item['metadata']['namespace']
-            name = item['metadata']['name']
+        if mismatched_secrets:
+            hookenv.log(
+                'ServiceAccount secrets do not have correct ca.crt: '
+                + ','.join(
+                    secret['metadata']['name'] for secret in mismatched_secrets
+                )
+            )
+            hookenv.log('Waiting to retry restarting addons')
+            return
+
+        # Now restart the addons
+        for deployment in deployments:
+            kind = deployment['kind']
+            namespace = deployment['metadata']['namespace']
+            name = deployment['metadata']['name']
             hookenv.log('Restarting addon: %s %s %s' % (kind, namespace, name))
             cmd = [
                 'kubectl', 'rollout', 'restart',
@@ -2464,8 +2528,8 @@ def restart_addons():
                 kind + '/' + name
             ]
             check_call(cmd)
-    except CalledProcessError:
+
+        leader_set({'kubernetes-master-addons-restart-for-ca': None})
+    except Exception:
         hookenv.log(traceback.format_exc())
         hookenv.log('Waiting to retry restarting addons')
-        return
-    leader_set({'kubernetes-master-addons-restart-needed': None})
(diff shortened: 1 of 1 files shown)


Cynerva:
`tls_client.certs.changed` is currently handled by [kick_api_server](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/aa7065651343626bac2ca16bcf849f6e9e051ab5/reactive/kubernetes_master.py#L919-L925). The new `update_certs` handler will conflict with it.

I do think that `kick_api_server` doesn't do enough. Presumably we'd want to restart kube-controller-manager, too, given that it takes in the CA and cert/key as options.

We probably don't need to restart kube-scheduler, but I also don't see any real harm in doing it.

All that said, I recommend removing the `kick_api_server` handler in favor of the new `update_cert` one.

johnsca:
Updated per review.  As per the other PRs (copying here for completeness), this was tested manually on AWS by doing the following:

1) Deploy CDK with the four patched charms & docker subordinate, wait to settle
2) Run some test pod in the cluster (I used busybox and left processes running on the pod)
3) Add Vault to the cluster, wait to settle
4) Unseal Vault, wait to settle
5) Remove EasyRSA, wait to settle
6) Update kubectl config
7) Confirm pod is still functioning

Cynerva:
> 4. Unseal Vault, wait to settle

In testing, during this step, I found that there is a race that can leave kubernetes-master in a bad state where kube-apiserver is no longer able to connect to etcd.

The following sequence occurred:
1. `certificates-relation-changed` hook ran, resulting in new certs being written, and kube-apiserver being restarted as expected. However, [cert information used to connect to etcd](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/9b72c10d6c87cd694cfe478f00c45d42905167f8/reactive/kubernetes_master.py#L1576-L1584) was -not- updated during this time, as those come from the etcd relation (handled in [handle_etcd_relation](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/9b72c10d6c87cd694cfe478f00c45d42905167f8/reactive/kubernetes_master.py#L1509-L1521)), not the certificates relation.
2. `etcd-relation-changed` hook ran, in which the etcd relation provided new client cert and CA info. However, start_master did not run again, so handle_etcd_relation was not called again, so the new certs were never written.

In this state, all kubectl calls fail. Hooks on kubernetes-master and kubernetes-worker are affected (staying in "executing" state for a long time, as they retry kubectl commands), and the cluster has not recovered on its own. Removing easyrsa did not help.

Cynerva:
Flannel also failed to switch to the new etcd client cert info. From `journalctl -o cat -u flannel`:
```
E0701 19:17:33.771678   29435 watch.go:171] Subnet watch failed: client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate signed by unknown authority
; error #1: x509: certificate signed by unknown authority
; error #2: x509: certificate signed by unknown authority
E0701 19:17:34.511649   29435 watch.go:43] Watch subnets: client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate signed by unknown authority
; error #1: x509: certificate signed by unknown authority
; error #2: x509: certificate signed by unknown authority
E0701 19:17:34.787105   29435 watch.go:171] Subnet watch failed: client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate signed by unknown authority
; error #1: x509: certificate signed by unknown authority
; error #2: x509: certificate signed by unknown authority
E0701 19:17:35.527465   29435 watch.go:43] Watch subnets: client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate signed by unknown authority
; error #1: x509: certificate signed by unknown authority
; error #2: x509: certificate signed by unknown authority
E0701 19:17:35.805219   29435 watch.go:171] Subnet watch failed: client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate signed by unknown authority
; error #1: x509: certificate signed by unknown authority
```

I suspect flannel, calico, canal, and tigera-secure-ee all need similar work to account for this transition.

johnsca:
@Cynerva Please take a look at the following PRs to address the issue in the other charms as well:

* https://github.com/charmed-kubernetes/charm-flannel/pull/55
* https://github.com/charmed-kubernetes/layer-calico/pull/38
* https://github.com/charmed-kubernetes/layer-canal/pull/35
* https://github.com/charmed-kubernetes/layer-tigera-secure-ee/pull/17

Cynerva:
Ready for review. Tested the latest commit as follows:

1. Deployed charmed-kubernetes with modified charms ([bundle.yaml](https://gist.github.com/Cynerva/9c4af6985e302d4131c553c9c2dc95c6)).
2. Deployed vault in auto-unlock mode:
```
juju deploy cs:percona-cluster
juju deploy cs:~openstack-charmers-next/vault
juju config vault auto-generate-root-ca-cert=true totally-unsecure-auto-unlock=true
juju relate vault percona-cluster
```
3. Waited for cluster to settle.
4. Related vault to etcd:
```
juju relate vault:certificates etcd
```
5. Waited for cluster to settle (took ~5 minutes).
6. Related vault to kubeapi-load-balancer, kubernetes-master, kubernetes-worker:
```
juju relate vault:certificates kubeapi-load-balancer
juju relate vault:certificates kubernetes-master
juju relate vault:certificates kubernetes-worker
```
7. Waited for cluster to settle (took another ~5 minutes).
8. Removed easyrsa:
```
juju remove-application easyrsa
```

Some comments:
1. Addons that needed restarting: coredns, metrics-server, heapster, nginx-ingress-controller. I suspect others likely need it too, e.g. csi-rbdplugin. To be safe, I made the charm restart all addons that are managed by CDK. We will need to document for users that they may need to manually restart addons they have brought themselves.
2. DNS impact is small, as CoreDNS still responds to DNS lookups while it is unable to reach k8s. Since the restart of CoreDNS is a rolling restart, there is no downtime.
3. Ingress impact is moderate, as ingress is temporarily unavailable while nginx-ingress-controller is restarted (because it is a DaemonSet that binds to host ports).
4. Doing the migration in two steps (etcd first, then kubernetes) reduces overall downtime, and in particular, seems to cause much less downtime for ingress.

Cynerva:
This now depends on https://github.com/charmed-kubernetes/cdk-addons/pull/133

I updated the addon restart code to select on a new `cdk-restart-on-ca-change` label.

Tested again as described in https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/34#issuecomment-510186591, went well. Needs review.

reactive/kubernetes_master.py
```diff
@@ -1853,9 +1867,14 @@ def token_generator(length=32):
...
 
-@retry(times=3, delay_secs=10)
+@retry(times=3, delay_secs=1)
 def get_pods(namespace='default'):
-    cmd = ['kubectl', 'get', 'po', '-n', namespace, '-o', 'json']
+    cmd = [
+        'kubectl', 'get', 'po',
+        '-n', namespace,
+        '-o', 'json',
+        '--request-timeout', '10s'
```
hyperbolic2346: 1) 10s seems extreme. Do we need to be that conservative?
2) Should this use the kubectl function in kubernetes_common?

reactive/kubernetes_master.py
```diff
@@ -2418,3 +2437,96 @@ def send_registry_location():
...
+            secret = json.loads(output)
+            secrets.append(secret)
+
+        # Check secrets have updated CA
+        with open(ca_crt_path, 'rb') as f:
+            ca = f.read()
+        encoded_ca = base64.b64encode(ca).decode('UTF-8')
+        mismatched_secrets = [
+            secret for secret in secrets
+            if secret['data']['ca.crt'] != encoded_ca
```
hyperbolic2346: Will this trip up if there are secrets named ca.cert in the cluster that aren't meant to be the same as this encoded ca?

hyperbolic2346: Nevermind, I see the selector on the query. We'll just need to be careful not to tag something that uses a different ca.

reactive/kubernetes_master.py
```diff
@@ -719,6 +721,21 @@ def start_master():
     hookenv.open_port(6443)
 
 
+@when('tls_client.certs.changed')
+def certs_changed():
+    remove_state('kubernetes-master.components.started')
```
hyperbolic2346: clear_flag?

reactive/kubernetes_master.py
Cynerva: 1. Can you clarify what you're wanting here? Is 10s too long or too short?

The default is 1m. I lowered it to 10s to reduce the amount of time kubernetes-master waits for futile kubectl calls during the migration. The apiserver can't reach etcd until kubernetes-master processes the `etcd-relation-changed` hook, and this was holding it up.

2. Sure, may as well.

Cynerva:
Do not merge, needs retest.

Updated kubectl calls in this PR to use `charms.layer.kubernetes_common.kubectl`, and changed the new `remove_state` calls to `clear_flag`. Needs review.

reactive/kubernetes_master.py
hyperbolic2346: My thought is that even 1 second seems way overkill in a local network. If this will connect to an apiserver that comes up within 10 seconds I could see keeping it this long. One minute was extreme and 10 seconds seems too long. I'm not strongly opposed to it, I just thought I would comment.

reactive/kubernetes_master.py
```diff
@@ -1869,14 +1870,13 @@ def token_generator(length=32):
...
-        '--request-timeout', '10s'
-    ]
     try:
-        output = check_output(cmd).decode('utf-8')
+        output = kubectl(
+            'get', 'po',
+            '-n', namespace,
+            '-o', 'json',
+            '--request-timeout', '10s'
+        ).decode('UTF-8')
```
hyperbolic2346: I don't believe the decodes are necessary here. Are you sure you need them?

hyperbolic2346: I don't see them in the kubectl helper, but I've used them to capture output without the decode in another place. We don't capture output in another spot in master or worker though.

Cynerva: Looks like it's not necessary in python 3.6

```
$ python3
Python 3.6.7 (default, Oct 31 2018, 13:29:56) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> import subprocess
>>> output = subprocess.check_output(['cat', 'some.json'])
>>> json.loads(output)
{'foo': 'bar'}
```

But it is necessary in python 3.5:
```
$ python3.5
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> import subprocess
>>> output = subprocess.check_output(['cat', 'some.json'])
>>> json.loads(output)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
>>> json.loads(output.decode('UTF-8'))
{'foo': 'bar'}
```

I don't have a full understanding of which python versions are used by the charms in what situations. My preference would be to play it safe and support both.

reactive/kubernetes_master.py
Cynerva: Thanks. I'd like to be cautious and err on the side of leaving it too long. Even with 10s here, most of kubernetes-master's time is spent elsewhere during the transition.

Cynerva:
Done testing, awaiting approval.

reactive/kubernetes_master.py
```diff
@@ -2466,7 +2466,7 @@ def restart_addons_for_ca():
         service_accounts = []
         for namespace, name in service_account_names:
             output = kubectl(
-                'kubectl', 'get', 'ServiceAccount', name,
+                'get', 'ServiceAccount', name,
```
hyperbolic2346: totally read over that, good catch.

PR #38, opened 2019-07-22
WIP fix missing system:nodes->system:node binding and add namespace:get to policy

joedborg:
Fixing https://bugs.launchpad.net/charm-kubernetes-master/+bug/1837249


joedborg:
Currently, this will get stuck on "Waiting for kube-system pods to start", so reactive won't execute the `apply` (you can jump in and do it manually).  Need to figure out how to put this ahead of the blocker.

joedborg:
[Cannot deploy pods after enabling RBAC](https://trello.com/c/tjDsWQdu/1319-cannot-deploy-pods-after-enabling-rbac)

templates/rbac-clusterrole.yaml
```diff
@@ -0,0 +1,188 @@
+# We overwrite system:node, rather than patching
+# due to a race condition where certain services
+# are expecting RBAC to be ready.
```
Cynerva: Is this necessary? I would expect to be able to create a new ClusterRole (possibly with a different name) that just grants "get namespace", and bind it to the system:nodes group, without clobbering existing permissions.

I'll look into this. Let me brush up on my understanding of RBAC and I'll get back to you.

joedborg: That wasn't working for me when trying; it'd overwrite `system:node`.  It's more than likely I was doing something wrong though :)

Cynerva: This worked for me: https://gist.github.com/Cynerva/5b5389581069c9e8ae9641acc490cb24

The different ClusterRole name (`namespace-reader` instead of `system:node`) is key there. This allows us to grant the "get namespace" permission via a secondary ClusterRole, without redefining the existing `system:node` one.

I think it's important that we do this. Overriding the `system:nodes` ClusterRole with our own definition may give us trouble down the line, if upstream k8s changes the default set of permissions for `system:node` but our charm overrides it with an older set.

Cynerva:
The "get namespace" requirement is specific to the Calico CNI plugin. I think it would be better to have the calico/canal/tigera-secure-ee charms apply this, rather than kubernetes-master.

templates/rbac-clusterrole.yaml
joedborg: Ah cool, I thought that it was required to be named `system:node` as well.  That's a much nicer fix.

joedborg:
@Cynerva have removed the Clusterrole from this PR and will move to the Calico one.

Cynerva:
Thanks. I think the ClusterRoleBinding should go over, too. Basically this PR goes away and we just add the ClusterRole and ClusterRoleBinding from https://gist.github.com/Cynerva/5b5389581069c9e8ae9641acc490cb24 to the Calico charm's templates instead.

joedborg:
All moved to https://github.com/charmed-kubernetes/layer-calico/pull/43

PR #39, opened 2019-07-23
Minor fixes following test on rbac-namespaces-fix

phvalguima:
* kubectl_manifest was looking into wrong place on create_rbac_namespaces
* Set switch_auth_mode to set flags during install_snaps; that avoids waiting for pods to start deadlock as described on PR


joedborg:
[Cannot deploy pods after enabling RBAC](https://trello.com/c/tjDsWQdu/1319-cannot-deploy-pods-after-enabling-rbac)

joedborg:
These have all been addressed in the CNI charms.  Thanks!

PR #43, opened 2019-08-01
Trigger reconfiguration of addons and api server when enable-metrics has been changed

joedborg:
Addresses https://bugs.launchpad.net/charm-kubernetes-master/+bug/1836076

Have tested on CK 1.15 stable.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -326,6 +326,28 @@ def install_snaps():
     remove_state('kubernetes-master.components.started')
 
 
+@when('etcd.available')
+@when('config.changed.enable-metrics')
+def enable_metric_changed():
+    """
+    Trigger an api server update.
+
+    :return: None
+    """
+    etcd = endpoint_from_flag('etcd.available')
+    connection = etcd.get_connection_string()
+    if not connection:
+        # etcd is not returning a connection string. This happens when
+        # the master unit disconnects from etcd and is ready to terminate.
+        # No point in trying to start master services and fail. Just return.
+        return
+
+    configure_apiserver(connection)
+
+    if is_state('leadership.is_leader'):
+        configure_cdk_addons()
+
+
 @when('config.changed.client_password', 'leadership.is_leader')
 def password_changed():
     """Handle password change via the charms config."""


joedborg:
[Enable metrics does not trigger hook](https://trello.com/c/P1Y6oz2I/1336-enable-metrics-does-not-trigger-hook)

reactive/kubernetes_master.py
```diff
@@ -326,6 +326,28 @@ def install_snaps():
...
+def enable_metric_changed():
+    """
+    Trigger an api server update.
+
+    :return: None
+    """
+    etcd = endpoint_from_flag('etcd.available')
+    connection = etcd.get_connection_string()
+    if not connection:
+        # etcd is not returning a connection string. This happens when
```
hyperbolic2346: I am concerned that we call configure_apiserver in a LOT of places in the code, but none of them have this check. Should the check be inside configure_apiserver instead so it catches it anywhere? I feel like we need to make a pass at unifying this and having an official way to handle reconfiguring. Maybe a flag that is cleared when it needs to happen. Seems that this code is copied all around and others may have attempted to start this with functions like reconfigure_cdk_addons.

reactive/kubernetes_master.py
```diff
@@ -326,6 +326,20 @@ def install_snaps():
...
 
+@when('etcd.available')
+@when('config.changed.enable-metrics')
+def enable_metric_changed():
+    """
+    Trigger an api server update.
+
+    :return: None
+    """
+    configure_apiserver(connection)
```
hyperbolic2346: What is connection? Where does it come from here?

reactive/kubernetes_master.py
```diff
@@ -326,6 +326,20 @@ def install_snaps():
     remove_state('kubernetes-master.components.started')
 
 
+@when('etcd.available')
+@when('config.changed.enable-metrics')
+def enable_metric_changed():
```
hyperbolic2346: Do we want to add a @when('kubernetes-master.components.started') here? If we get config changed before things are started we will attempt to configure the apiserver and cdk-addons before they are ready. I'm not sure if this would cause trouble, but could it be before the snap is installed? On the other hand, if we ignore a config changed if the components aren't started, starting the components pretty much runs this function, so it will start correctly the first time, right?

PR #45, opened 2019-08-06
Add support for new Prometheus & Grafana relations

johnsca:
Instead of requiring the operator to manually find, download, and provide the scraper job to Prometheus and dashboards to Grafana, this uses the new relations to have the k8s-master charm provide them automatically.

Depends on:
  * https://github.com/juju-solutions/charms.reactive/pull/215
  * https://github.com/juju-solutions/interface-prometheus-manual/pull/1
  * [lp:~johnsca/prometheus2-charm#371005](https://code.launchpad.net/~johnsca/prometheus2-charm/+git/prometheus2-charm/+merge/371005)
  * https://github.com/juju-solutions/interface-grafana-dashboard/pull/1
  * [lp:~johnsca/grafana-charm#371002](https://code.launchpad.net/~johnsca/grafana-charm/+git/grafana-charm/+merge/371002)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -977,7 +977,8 @@ def configure_cdk_addons():
     # when configuring the cdk-addons snap until 1.17 is released.
     registry = hookenv.config('addons-registry')
     if registry and get_version('kube-apiserver') < (1, 17):
-        hookenv.log('addons-registry is deprecated; use image-registry instead')
+        hookenv.log('addons-registry is deprecated; '
+                    'use image-registry instead')
     else:
         registry = hookenv.config('image-registry')
     dbEnabled = str(hookenv.config('enable-dashboard-addons')).lower()
@@ -2575,3 +2576,77 @@ def add_systemd_iptables_patch():
 
     # enable and run the service
     service_resume(service_name)
+
+
+@when('leadership.is_leader',
+      'kubernetes-master.components.started',
+      'endpoint.prometheus.joined',
+      'certificates.ca.available')
+def register_prometheus_jobs():
+    prometheus = endpoint_from_flag('endpoint.prometheus.joined')
+    # tls = endpoint_from_flag('certificates.ca.available')
+    loadbalancer = endpoint_from_flag('loadbalancer.available')
+    forced_lb_ips = hookenv.config('loadbalancer-ips').split()
+    hacluster_vip = get_hacluster_ip_or_hostname()
+
+    if forced_lb_ips:
+        # if the user gave us IPs for the load balancer, assume they know
+        # what they are talking about and use that instead of our information.
+        round_robin = get_unit_number() % len(forced_lb_ips)
+        k8s_api_endpoint = forced_lb_ips[round_robin]
+    elif hacluster_vip:
+        k8s_api_endpoint = hacluster_vip
+    elif loadbalancer:
+        lb_addresses = loadbalancer.get_addresses_ports()
+        k8s_api_endpoint = lb_addresses[0].get('public-address')
+    else:
+        k8s_api_endpoint = prometheus.ingress_address
+    k8s_password = get_password('basic_auth.csv', 'admin')
+
+    templates_dir = Path('templates')
+    for job_file in Path('templates/prometheus').glob('*.yaml.j2'):
+        prometheus.register_job(
+            job_name=job_file.name.split('.')[0],
+            job_data=yaml.safe_load(
+                render(source=job_file.relative_to(templates_dir),
+                       target=None,  # don't write file, just return rendered
+                       context={
+                           'k8s_api_endpoint': k8s_api_endpoint,
+                           'k8s_password': k8s_password,
+                       })
+            ),
+        )
+
+
+def detect_telegraf():
+    # Telegraf uses the implicit juju-info relation, which makes it difficult
+    # to tell if it's related. The "best" option is to look for the subordinate
+    # charm on disk.
+    for charm_dir in Path('/var/lib/juju/agents').glob('unit-*/charm'):
+        metadata = yaml.safe_load((charm_dir / 'metadata.yaml').read_text())
+        if 'telegraf' in metadata['name']:
+            return True
+    else:
+        return False
+
+
+@when('leadership.is_leader',
+      'kubernetes-master.components.started',
+      'endpoint.grafana.joined')
+def register_grafana_dashboards():
+    grafana = endpoint_from_flag('endpoint.grafana.joined')
+
+    # load conditional dashboards
+    dash_dir = Path('templates/grafana/conditional')
+    if is_flag_set('endpoint.prometheus.joined'):
+        dashboard = (dash_dir / 'prometheus.json').read_text()
+        grafana.register_dashboard('prometheus', json.loads(dashboard))
+    if detect_telegraf():
+        dashboard = (dash_dir / 'telegraf.json').read_text()
+        grafana.register_dashboard('telegraf', json.loads(dashboard))
+
+    # load automatic dashboards
+    dash_dir = Path('templates/grafana/autoload')
+    for dash_file in dash_dir.glob('*.json'):
+        dashboard = dash_file.read_text()
+        grafana.register_dashboard('telegraf', json.loads(dashboard))
--- a/templates/prometheus/k8s-api-endpoints.yaml.j2
+++ b/templates/prometheus/k8s-api-endpoints.yaml.j2
@@ -1,0 +1,20 @@
+job_name: 'k8s-api-endpoints'
+kubernetes_sd_configs:
+- api_server: https://{{k8s_api_endpoint}}
+  role: endpoints
+  tls_config:
+    insecure_skip_verify: true
+  basic_auth:
+    username: admin
+    password: {{k8s_password}}
+scrape_interval: 30s
+scheme: https
+tls_config:
+  insecure_skip_verify: true
+basic_auth:
+  username: admin
+  password: {{k8s_password}}
+relabel_configs:
+- source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
+  action: keep
+  regex: default;kubernetes;https
--- a/templates/prometheus/kubernetes-nodes.yaml.j2
+++ b/templates/prometheus/kubernetes-nodes.yaml.j2
@@ -1,0 +1,25 @@
+job_name: 'kubernetes-nodes'
+kubernetes_sd_configs:
+- api_server: https://{{k8s_api_endpoint}}
+  role: node
+  tls_config:
+    insecure_skip_verify: true
+  basic_auth:
+    username: admin
+    password: {{k8s_password}}
+scrape_interval: 30s
+scheme: https
+tls_config:
+  insecure_skip_verify: true
+basic_auth:
+  username: admin
+  password: {{k8s_password}}
+relabel_configs:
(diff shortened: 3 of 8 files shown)


reactive/kubernetes_master.py
```diff
@@ -2575,3 +2576,77 @@ def add_systemd_iptables_patch():
...
     service_resume(service_name)
+
+
+@when('leadership.is_leader',
+      'kubernetes-master.components.started',
+      'endpoint.prometheus.joined',
+      'certificates.ca.available')
+def register_prometheus_jobs():
+    prometheus = endpoint_from_flag('endpoint.prometheus.joined')
+    # tls = endpoint_from_flag('certificates.ca.available')
```
hyperbolic2346: remove?

reactive/kubernetes_master.py
```diff
@@ -2575,3 +2576,77 @@ def add_systemd_iptables_patch():
...
+    loadbalancer = endpoint_from_flag('loadbalancer.available')
+    forced_lb_ips = hookenv.config('loadbalancer-ips').split()
+    hacluster_vip = get_hacluster_ip_or_hostname()
+
+    if forced_lb_ips:
```
hyperbolic2346: I wonder if we're to the point of moving this into a function. This is copied from somewhere else in here right? The logic for "get my api server endpoint" is getting complex and I would hate to end up with a bug only in one path.

reactive/kubernetes_master.py
```diff
@@ -2575,3 +2576,77 @@ def add_systemd_iptables_patch():
...
+    elif loadbalancer:
+        lb_addresses = loadbalancer.get_addresses_ports()
+        k8s_api_endpoint = lb_addresses[0].get('public-address')
+    else:
+        k8s_api_endpoint = prometheus.ingress_address
+    k8s_password = get_password('basic_auth.csv', 'admin')
+
+    templates_dir = Path('templates')
+    for job_file in Path('templates/prometheus').glob('*.yaml.j2'):
+        prometheus.register_job(
```
hyperbolic2346: Is it cool to register the jobs every hook run? Should we use a flag to only do this once?

reactive/kubernetes_master.py
```diff
@@ -2575,3 +2576,77 @@ def add_systemd_iptables_patch():
...
+                           'k8s_password': k8s_password,
+                       })
+            ),
+        )
+
+
+def detect_telegraf():
+    # Telegraf uses the implicit juju-info relation, which makes it difficult
+    # to tell if it's related. The "best" option is to look for the subordinate
+    # charm on disk.
```
hyperbolic2346: On DISK? This seems very brittle. :disappointed:

templates/prometheus/k8s-api-endpoints.yaml.j2
```diff
@@ -0,0 +1,20 @@
+job_name: 'k8s-api-endpoints'
+kubernetes_sd_configs:
+- api_server: https://{{k8s_api_endpoint}}
+  role: endpoints
+  tls_config:
+    insecure_skip_verify: true
```
hyperbolic2346: :open_mouth:

templates/prometheus/kubernetes-nodes.yaml.j2
```diff
@@ -0,0 +1,25 @@
+job_name: 'kubernetes-nodes'
+kubernetes_sd_configs:
+- api_server: https://{{k8s_api_endpoint}}
+  role: node
+  tls_config:
+    insecure_skip_verify: true
+  basic_auth:
+    username: admin
```
hyperbolic2346: Do we want to give the admin user creds out? Maybe a view-only user?

reactive/kubernetes_master.py
johnsca: This was actually a placeholder for improving the auth story that I forgot to circle back around to.  I'm going to go ahead and do that now.

reactive/kubernetes_master.py
johnsca: Yeah, I started to break this out into a function but then noticed that the code-paths were already slightly different and neither seemed exactly correct.  I think what I have here is the simplest and most correct of the 3(?), so if you're ok with that, I'll replace the others with this as a function.

reactive/kubernetes_master.py
johnsca: We do want to re-check the values, in case they change, and if they don't then the rendered template should be stable so it shouldn't cause any extra hooks to fire on the other end, but I can add explicit logic to watch for changes in the input values if you think that's cleaner (explicit over implicit).

reactive/kubernetes_master.py
johnsca: Yeah, it does.  I had discussed this with @kwmonroe and we couldn't come up with a better solution, but I'd really love to find one, if you have some suggestions.  Options we considered:

* Make the dashboard unconditional (the deployment might not even use Telegraf)
* Make the dashboard conditional on the presence of Prometheus (could use Prometheus w/o Telegraf)
* Make the dashboard conditional on `which telegraf` (still requires on-disk checks and is really just brittle in a different way)

Thrown into the mix is that the Grafana charm includes a Telegraf dashboard out of the box which has some overlap, but which apparently [might not always be loaded](https://github.com/conjure-up/spells/blob/master/charmed-kubernetes/addons/prometheus/steps/01_install-prometheus/after-deploy#L28-L31) when we want it to be.

templates/prometheus/k8s-api-endpoints.yaml.j2
johnsca: There is a separate bug about improving auth and security for this scraper, and I laid the groundwork for addressing both of those in the interface layers, but forgot to follow-up on it in this charm.  I'll take a stab at it.

templates/prometheus/kubernetes-nodes.yaml.j2
johnsca: See previous comments about the separate bug.  I'll take a stab at this but if it starts taking too long, I might have to leave it as a separate work item.

reactive/kubernetes_master.py
```diff
@@ -851,30 +852,8 @@ def push_service_data():
...
-            address = vips
-        elif dns_record:
-            address = dns_record
-
-    if address:
-        kube_api.configure(6443, address, address)
-    else:
-        kube_api.configure(6443)
+    address, port = kubernetes_master.get_api_endpoint(kube_api.relations[0])
+    kube_api.configure(port, address, address)
```
hyperbolic2346: Is this the same behavior as before? We only set address if it was forced or we had hacluster before. get_api_endpoint now sets it to hookenv.unit_public_address as a default. configure defaults to hookenv.ingress_address. Will this break spaces?

reactive/kubernetes_master.py
```diff
@@ -906,6 +885,9 @@ def send_data():
         'kubernetes.default.svc.{0}'.format(domain)
     ]
 
+    # NB: this section can't be refactored to use
```
hyperbolic2346: This makes me sad. Maybe we should refactor the function to return a list and have the places that don't want more than one do the selection based on unit? Or maybe an optional param to say "return everything"?
`def get_api_endpoints():
<contents of get_api_endpoint minus culling>

def get_api_endpoint():
  ips = get_api_endpoints()
  return ips[kubernetes_common.get_unit_number() % len(ips)]

reactive/kubernetes_master.py
Cynerva: @hyperbolic2346 beat me to it, but here it is anyway.

Grabbing `kube_api.relations[0]` here seems wrong. Before, in the non-`forced_lb_ips`, non-`hacluster` case, we would invoke `kube_api.configure(6443)` with no address and let the http interface fall back to the ingress address for each relation: https://github.com/juju-solutions/interface-http/blob/632131b1f122daf6fb601fd4c9f1e4dbb1a92e09/provides.py#L53-L54.

With this change, in the case described above, every relation will be configured with the ingress address of the -first- relation.

Isn't it possible for each relation to have a different ingress address? I would expect this to break in scenarios involving multiple network spaces or cross-model relations.

reactive/kubernetes_master.py
```diff
@@ -1487,10 +1422,15 @@ def shutdown():
         service_stop('snap.%s.daemon' % service)
 
 
-def build_kubeconfig(server):
+@when('certificates.ca.available',
```
hyperbolic2346: :+1:

reactive/kubernetes_master.py
```diff
@@ -2585,22 +2507,8 @@ def add_systemd_iptables_patch():
...
-        round_robin = get_unit_number() % len(forced_lb_ips)
-        k8s_api_endpoint = forced_lb_ips[round_robin]
-    elif hacluster_vip:
-        k8s_api_endpoint = hacluster_vip
-    elif loadbalancer:
-        lb_addresses = loadbalancer.get_addresses_ports()
-        k8s_api_endpoint = lb_addresses[0].get('public-address')
-    else:
-        k8s_api_endpoint = prometheus.ingress_address
+    address, port = kubernetes_master.get_api_endpoint(prometheus.relations[0])
```
hyperbolic2346: :tada:

reactive/kubernetes_master.py
Cynerva: Even worse, it looks like when the loadbalancer relation is connected, kubernetes-master will send the loadbalancer's address over the kube-api-endpoint relation instead of its own. The kubeapi-load-balancer charm uses info from that relation to determine where to proxy traffic to the masters, but the masters will be telling it to proxy traffic to itself.

Cynerva: Also, in the forced_lb_ips and hacluster cases, this would previously send a list of addresses to the http relation. Now it only sends one.

johnsca: I was assuming a single LB relation, but you're right that we shouldn't make that assumption if we can avoid it.  As for sending the LB address, I read the comment that was removed to mean that the value would be ignored if the LB was connected, but I see that I probably understood that wrong.  I think I'll have to exclude this one from the refactor as well, meaning it will only be touching I think one other place.  Frustrating that this logic is slightly different in each place making it difficult to factor up.

reactive/kubernetes_master.py
hyperbolic2346: :crying_cat_face:

lib/charms/layer/kubernetes_master.py
```diff
@@ -0,0 +1,65 @@
...
+    dns_record = hookenv.config('ha-cluster-dns')
+    if forced_lb_ips:
+        # if the user gave us IPs for the load balancer, assume
+        # they know what they are talking about and use that
+        # instead of our information.
+        return [(address, STANDARD_API_PORT) for address in forced_lb_ips]
+    elif vips:
+        return [(vip, STANDARD_API_PORT) for vip in vips]
+    elif dns_record:
+        return [(dns_record, STANDARD_API_PORT)]
```
Cynerva: Previous behavior was to not use ha-cluster-vip or ha-cluster-dns unless the ha.connected flag is set. Looks like that was lost somewhere in the refactor.

Cynerva:
I like the refactor. Thanks. The new prometheus/grafana bits look reasonable to me as well. https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/45#discussion_r311775572 is my last concern but this otherwise looks good to me.

Cynerva:
:+1: LGTM

johnsca:
Tested on AWS with only kube-api-loadbalancer and grafana exposed and both dashboards loaded and had data.

PR #46, opened 2019-08-22
Controller manager on secured port

pierrop:
Hi,

**Motivation for this PR**:
At this moment controller managers services are talking to API server on the insecured port. (i.e:: auth and authz bypassed) PodSecurityPolicy need controller-manager service running, authentificated with it's own credential. Otherwise all PodSecurityPolicy objects would be allowed for each Pod created by controller-manager (ReplicaSet, Deployment...)

I have a fork of this charm with this change since more than 6 months and I'm interested to push it upstream.

Thanks,

PS: for the variable "kubecontrollermanagerconfig_path" maybe it's better to have it in kubernetes_common layer. If this PR is accepted I will create another PR in kubernetes_common to put this variable in the right place by refering this PR

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1805,7 +1805,10 @@ def configure_controller_manager():
     controller_opts['v'] = '2'
     controller_opts['root-ca-file'] = str(ca_crt_path)
     controller_opts['logtostderr'] = 'true'
-    controller_opts['master'] = 'http://127.0.0.1:8080'
+    controller_opts['kubeconfig'] = kubecontrollermanagerconfig_path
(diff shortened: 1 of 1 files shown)


hyperbolic2346:
Thank you very much for the PR, @pierrop. I wouldn't suggest moving kubecontrollermanagerconfig_path to kubernetes_common unless it was needed by the master and worker charm.

Cynerva:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1841965

Thanks for the follow-up @pierrop. LGTM, merging!

pierrop:
Perfect ! You're welcome :)

PR #51, opened 2019-08-30
Revert enabling non-OpenStack cloud provider settings in cdk-addons

johnsca:
Until we support the other external cloud providers, it's safer to leave the flags set to False.  This means the storage class won't be automatically created, but it also means that the transition to the external cloud provider will be seamless when the time comes.


johnsca:
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1842125

PR #52, opened 2019-08-31
Support volume config for grafana/influxdb

xtrusia:
Adding configuration to setup volume for grafana/influxdb

Closes-bug:#1830118


Cynerva:
LGTM, thanks!

PR #53, opened 2019-09-10
Remove direct call to reconfigure

hyperbolic2346:
The preferred method is now via flag removal. Removing api server config flag to force a reconfigure. Note that the timing of this isn’t critical, so I am not concerned about synchronous call vs asynchronous.


Cynerva:
LGTM, but this needs to be tied back to a launchpad bug

hyperbolic2346:
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1843506

PR #54, opened 2019-09-11
Add cluster-name param to CM and addons

johnsca:
Load balancers created by Kubernetes are prefixed by the cluster name, which defaults to "kubernetes". If this is not overridden, then load balancers (and potentially other resources) from different clusters in the same cloud can conflict.  This uses the cluster tag we already generate as the cluster name (and passes it through to cdk-addons for the external OpenStack provider) to ensure uniqueness and consistency with the way we annotate resources belonging to this cluster.

Fixes [lp:1842941](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1842941)
Depends on https://github.com/charmed-kubernetes/cdk-addons/pull/147


johnsca:
Tested with Octavia-enabled cloud, and verified K8s-created LBs include the UUID to ensure uniqueness.

PR #57, opened 2019-09-18
Prevent cluster from starting when Vault is pending

johnsca:
When Vault is related to provide encryption-at-rest, we should not start the cluster and potentially allow unencrypted secrets to be created while we're still waiting for Vault.

Related to: [lp:1843809](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1843809)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -500,6 +500,22 @@ def set_app_version():
     hookenv.application_version_set(version.split(b' v')[-1].rstrip())
 
 
+@hookenv.atstart
+def check_vault_pending():
+    try:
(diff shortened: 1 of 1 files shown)


johnsca:
Tested on AWS along with:
* https://github.com/openstack-charmers/charm-interface-vault-kv/pull/6
* https://github.com/openstack-charmers/charm-interface-vault-kv/pull/7
* https://github.com/juju-solutions/layer-vault-kv/pull/7

Confirmed that the cluster blocked until Vault was manually unsealed, then successfully came up.  Then confirmed that the relation was able to be removed without error.

PR #63, opened 2019-10-02
Adding nagios plugin to monitor the API server.

hyperbolic2346:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1844549 and https://bugs.launchpad.net/charm-kubernetes-master/+bug/1844547

This plugin verifies the API server is reachable on the localhost and the external IP. It also verifies the snaps are installed.

Nagios reports on crash-looping services as tested by adding invalid command-line options for the API server and the controller manager. Both resulted in errors reported by nagios.

This change depends on a nagios-layer change and can not be merged before that is merged: https://github.com/charmed-kubernetes/layer-nagios/pull/1

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1294,9 +1294,9 @@ def ceph_storage():
 
 @when('nrpe-external-master.available')
 @when_not('nrpe-external-master.initial-config')
-def initial_nrpe_config(nagios=None):
+def initial_nrpe_config():
     set_state('nrpe-external-master.initial-config')
-    update_nrpe_config(nagios)
+    update_nrpe_config()
 
 
 @when('config.changed.authorization-mode')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -88,7 +88,8 @@ from charms.layer.kubernetes_common import client_crt_path
 from charms.layer.kubernetes_common import client_key_path
 from charms.layer.kubernetes_common import kubectl
 
-from charms.layer.nagios import install_nagios_plugin, remove_nagios_plugin
+from charms.layer.nagios import install_nagios_plugin_from_file
+from charms.layer.nagios import remove_nagios_plugin
 
 
 # Override the default nagios shortname regex to allow periods, which we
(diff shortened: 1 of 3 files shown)


templates/nagios_plugin.py
```diff
@@ -0,0 +1,58 @@
...
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(1)
+        s.connect((host, int(port)))
+        s.shutdown(socket.SHUT_RDWR)
+    finally:
+        s.close()
+
+
+def verify_local_connection_to_apiserver():
```
tvansteenburgh: Not sure we should include this since we'll likely be shutting down the insecure local port soon, to conform to the CIS benchmark for k8s. Check me on that @kwmonroe. I think the remote 6443 check is sufficient.

tvansteenburgh:
You can add a new mock to https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/tests/conftest.py to fix the test failure.

templates/nagios_plugin.py
kwmonroe: Correct, it'll be a tech preview to disable 8080 for CIS in 1.17, with total removal of the insecure port in a subsequent release.

reactive/kubernetes_master.py
```diff
@@ -1361,30 +1367,30 @@ def remove_rbac_resources():
...
+
+    plugin = install_nagios_plugin_from_file('templates/nagios_plugin.py',
+                                             'check_k8s_master.py')
     hostname = nrpe.get_nagios_hostname()
     current_unit = nrpe.get_nagios_unit_name()
     nrpe_setup = nrpe.NRPE(hostname=hostname)
     nrpe.add_init_service_checks(nrpe_setup, services, current_unit)
+    nrpe_setup.add_check('k8s-api-server',
+                         'Verify that the Kubernetes API server is accessible',
+                         str(plugin))
```
kwmonroe: can we call this check `k8s-master` instead of `k8s-api-server`?  we're checking 3 services on the same machine here, so it seems weird to umbrella them with `api-server` when we actually care about the c-manager and scheduler too.

reactive/kubernetes_master.py
```diff
@@ -1393,6 +1399,7 @@ def remove_nrpe_config(nagios=None):
 
     for service in services:
         nrpe_setup.remove_check(shortname=service)
+    nrpe_setup.remove_check(shortname='k8s-api-server')
```
kwmonroe: if we change this to `k8s-master`, be sure to update this to match.

reactive/kubernetes_master.py
```diff
@@ -1361,30 +1367,30 @@ def remove_rbac_resources():
...
+                         'Verify that the Kubernetes API server is accessible',
+                         str(plugin))
     nrpe_setup.write()
 
 
 @when_not('nrpe-external-master.available')
 @when('nrpe-external-master.initial-config')
-def remove_nrpe_config(nagios=None):
+def remove_nrpe_config():
     remove_state('nrpe-external-master.initial-config')
```
kwmonroe: let's remove this state at the end.  otherwise, we could fail in this handler and incorrectly assume we've cleaned up.

reactive/kubernetes_master.py
hyperbolic2346: This specific plugin verifies that all snaps are installed and then verifies that it can talk to the API server. It is a bit of a misnomer to call it just apiserver, but it isn't checking that the other services are up. There are other nagios checks for that.

reactive/kubernetes_master.py
hyperbolic2346: yes, I'd love to change existing code on a merged PR. :)

That said, this seems reasonable, want me to make a new PR for that?

kwmonroe: Done in #65

PR #66, opened 2019-10-29
Handle upgrades for sending new cluster-name to cdk-addons

johnsca:
This fixes the issue where upgrading the charm code will cause the OpenStack cloud provider to use a new `cluster-name` value and thus recreate all load balancers with new names, requiring manual intervention. This makes it so the operator needs to invoke the upgrade action to enable the new cluster name and can explicitly prevent the switch by overriding the `fix-cluster-name` action param.

Fixes [lp:1850539](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1850539)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -180,10 +180,26 @@ def maybe_install_kube_proxy():
         calculate_and_store_resource_checksums(checksum_prefix, snap_resources)
 
 
+@when_not('kubernetes-master.first-init.ran')
+def first_init():
+    # Do anything that we want to only do once.
+    if not is_state('kubernetes-master.cluster-name.legacy'):
+        # For new installs, this will run before check_for_upgrade_needed,
+        # but for upgrades from older code, that will run first.
+        set_state('kubernetes-master.cluster-name.new')
+    set_state('kubernetes-master.first-init.ran')
+
+
 @hook('upgrade-charm')
 def check_for_upgrade_needed():
     '''An upgrade charm event was triggered by Juju, react to that here.'''
     hookenv.status_set('maintenance', 'Checking resources')
+
+    # Handle upgrade path for cluster name being sent to cdk-addons.
+    if not is_state('kubernetes-master.cluster-name.new'):
+        # For new installs, first_init will run before this, but for
+        # upgrades from older code, this will run first.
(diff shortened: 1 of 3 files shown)


reactive/kubernetes_master.py
```diff
@@ -180,11 +180,27 @@ def maybe_install_kube_proxy():
...
+
 @hook('upgrade-charm')
 def check_for_upgrade_needed():
     '''An upgrade charm event was triggered by Juju, react to that here.'''
     hookenv.status_set('maintenance', 'Checking resources')
 
+    # Handle upgrade path for cluster name being sent to cdk-addons.
+    if not is_state('kubernetes-master.cluster-name.new'):
+        # For new installs, first_init will run before this, but for
+        # upgrades from older code, this will run first.
```
kwmonroe: You sure `check_for_upgrade_needed` will run before `first_init`?  Does @hook always happen before reactive handlers?

reactive/kubernetes_master.py
```diff
@@ -1084,6 +1100,10 @@ def configure_cdk_addons():
     enable_gcp = 'false'
     enable_openstack = str(is_flag_set('endpoint.openstack.ready')).lower()
     openstack = endpoint_from_flag('endpoint.openstack.ready')
+    if is_state('kubernetes-master.cluster-name.new'):
+        cluster_name = leader_get('cluster_tag')
+    elif is_state('kubernetes-master.cluster-name.legacy'):
```
kwmonroe: Do we really need two flags here?  Why not just react to `if .new: foo; else: bar`?

reactive/kubernetes_master.py
johnsca: Yes, `@hook` always runs before any `@when` handlers.  Still, I'm not super happy with this, as it feels too implicit / fragile.  But I couldn't think of another way to do it.

reactive/kubernetes_master.py
johnsca: We need two flags because we need to handle four cases: a fresh install, an upgrade of old code that hasn't called the `upgrade` action with the flag yet, an upgrade of old code which has already called the action, and an upgrade of new code.  I couldn't figure out another way to distinguish between those with a single flag.  (Some of those cases could maybe be combined, but I still ended up needing both flags.)

PR #68, opened 2019-11-01
Adding retries to keystone policy file changes.

hyperbolic2346:
Also fixed that policy file wouldn't be applied/deleted when toggling authorization.
Added a message for the user to know things were not ready/idle when the policy
file wasn't applied yet.

Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1850968


Cynerva:
> Also fixed that policy file wouldn't be applied/deleted when toggling authorization.

I'm confused. Why does toggling authorization mean the policy should be applied or deleted? If I'm reading right, the policy is needed with or without authorization because it's used to configure authentication, not authorization.

More specifically, it looks like what you've done is make it so when `config.changed.enable-keystone-authorization` is set, then `generate_keystone_configmap` will get run again. But `generate_keystone_configmap` doesn't seem to reference the `enable-keystone-authorization` config, so it looks like nothing would actually change. What am I missing?

> Added a message for the user to know things were not ready/idle when the policy
file wasn't applied yet.

Nice :+1: Thanks for that.

Cynerva:
Has this been tested?

hyperbolic2346:
> I'm confused. Why does toggling authorization mean the policy should be applied or deleted? If I'm reading right, the policy is needed with or without authorization because it's used to configure authentication, not authorization.

It's actually not for authentication, but instead authorization. The policy file maps from keystone users to RBAC rules. If you just have authentication turned on(default), then any keystone user can do what they want.

> More specifically, it looks like what you've done is make it so when `config.changed.enable-keystone-authorization` is set, then `generate_keystone_configmap` will get run again. But `generate_keystone_configmap` doesn't seem to reference the `enable-keystone-authorization` config, so it looks like nothing would actually change. What am I missing?

We need to apply this policy file when authorization is set and make sure it is gone when it is not set. You are correct that the current code isn't exactly what is wanted. I'm missing a check for authorization enabled before writing the policy file to the cluster. I have a fix for that incoming.

As for testing. This was tested, but I'm sure not sufficiently. Specifically, I tested authorization enabled, but not with authorization disabled or I would hopefully have noticed this error. I tested upgrading a cluster to this charm with keystone authentication and also spinning up a new cluster with this code. All done on AWS.

I'm adding to that testing ensuring that the authorization policy file is removed when upgrading and still using authentication only and also when disabling authorization. I'll reply again when that is completed.

Cynerva:
> It's actually not for authentication, but instead authorization.

Ahh okay, that makes sense, thanks. In that case, can you update the keystone-policy config description to say authorization instead of authentication?

https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/d58cd511ab72a371fe507d2cfa0e490419e6f3b1/config.yaml#L256

hyperbolic2346:
After testing and thinking about this for a while, I have changed this PR. It still accomplishes the original goal of retrying the keystone policy file apply, but I added some to it for robustness and to help the user.

 * Reworded the config to remove the word authentication from the policy file since it is for authorization.
 * Handle removing of the policy config map upon removing keystone relation
 * Surface error to the user if the policy file fails to apply or is empty
 * Retry applying policy file until success to prevent a restarting API server or something from breaking keystone.

My only concern is with setting an error flag if the `kubectl apply` fails. The error message says that the policy file is invalid, but it could be that the api server is restarting or something. I did add the error all the way at the end, so it should only surface if everything is running. I still worry it might pop up if the policy is changed at exactly the wrong time though.

Cynerva:
LGTM, thanks :+1:

PR #70, opened 2019-11-19
Add support for ClusterDNS k8s Charms

DomFleischmann:
This will add support for the cluster-dns interface which will allow
us to install DNS services like the CoreDNS k8s charm and let them
be integrated with Charmed Kubernetes.


Cynerva:
Looks like you accidentally reverted https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/61 in this PR. Can you fix that?

Cynerva:
Also, this needs to be linked to a Launchpad bug

layer.yaml
```diff
@@ -12,6 +12,7 @@ includes:
   - 'layer:vault-kv'
   - 'layer:vaultlocker'
   - 'layer:hacluster'
+  - 'interface:cluster-dns'
   - 'interface:ceph-admin'
   - 'interface:ceph-client'
   - 'interface:etcd'
```
Cynerva: Can you remove `interface:kube-dns` from this file while too?

metadata.yaml
```diff
@@ -29,11 +29,11 @@ peers:
 provides:
   kube-api-endpoint:
     interface: http
-  cluster-dns:
+  #cluster-dns:
     # kube-dns is deprecated. Its functionality has been rolled into the
     # kube-control interface. The cluster-dns relation will be removed in
     # a future release.
-    interface: kube-dns
+    #interface: kube-dns
```
Cynerva: Can you remove the above 5 lines, rather than commenting the 2 lines out?

This comment saying "The cluster-dns relation will be removed in a future release." could be confusing for future readers of this file.

layer.yaml
```diff
@@ -12,6 +12,7 @@ includes:
   - 'layer:vault-kv'
   - 'layer:vaultlocker'
   - 'layer:hacluster'
+  - 'interface:cluster-dns'
```
Cynerva: Make sure you add this new interface to the [layer index](https://github.com/juju/layer-index)

reactive/kubernetes_master.py
```diff
@@ -2818,3 +2819,32 @@ def api_server_stopped():
...
+def send_clusterdns_domain():
+    """Send domain to the Cluster DNS interface."""
+    cluster_dns = endpoint_from_flag('cluster-dns.joined')
+    dns_domain = hookenv.config('dns_domain')
+
+    hookenv.log('Sending dns_domain: {}'.format(dns_domain))
+    cluster_dns.send_domain(dns_domain)
+
+    set_flag('kubernetes-master.need-cluster-dns')
+    clear_flag('cluster-dns.joined')
```
Cynerva: What if dns_domain changes? i.e. user runs `juju config kubernetes-master dns_domain=new.domain`. This handler is set up to only run once, so the new value will never be sent, right?

reactive/kubernetes_master.py
```diff
@@ -2818,3 +2819,32 @@ def api_server_stopped():
...
+    """Send new ClusterDNS information to the worker"""
+    kube_control = endpoint_from_flag('kube-control.connected')
+    cluster_dns = endpoint_from_flag('cluster-dns.ready')
+    dns_ip = cluster_dns.get_ip()
+    dns_domain = hookenv.config('dns_domain')
+
+    hookenv.log('Sending dns_ip: {}'.format(dns_ip))
+    kube_control.set_dns(53, dns_domain, dns_ip, True)
+
+    clear_flag('kubernetes-master.need-cluster-dns')
```
Cynerva: What if `dns_domain` changes? What if `dns_ip` changes?

layer.yaml
DomFleischmann: [Done!](https://github.com/juju/layer-index/pull/101)

reactive/kubernetes_master.py
DomFleischmann: I will have to take a look at that

reactive/kubernetes_master.py
DomFleischmann: How could an IP of a service change without restarting the whole service?

Cynerva: No idea. It's generally not supposed to happen. But I've seen the occasional bug report where the kube-dns service IP changed unexpectedly and the user had to recreate all of their pods to pick up the new DNS IP. It's rare, and I have no idea how it happens, but it happens. Maybe it will never happen with the coredns k8s charm, but we should be prepared for that possibility and handle it appropriately.

DomFleischmann: I think juju would also get lost if the service underneath the k8s charm changes IP. But I will ask the juju k8s team.

johnsca:
Closing this in favor of the already landed #126

PR #74, opened 2020-02-11
Add support for CephFS

johnsca:
Adds detecting of CephFS when Ceph is related and, if present, informs cdk-addons to enable the cephfs-csi plugin.

Part of https://bugs.launchpad.net/charm-kubernetes-master/+bug/1862217

### DIFF
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -1,11 +1,12 @@
 import json
 from pathlib import Path
-from subprocess import check_call, check_output
+from subprocess import check_output
 from tempfile import TemporaryDirectory
 
 from charmhelpers.core import hookenv
 from charms.reactive import endpoint_from_flag, is_flag_set, set_flag
 
+from charms.layer import basic
 from charms.layer import kubernetes_common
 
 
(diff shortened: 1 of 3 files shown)


lib/charms/layer/kubernetes_master.py
```diff
@@ -64,3 +69,25 @@ def get_api_endpoint(relation=None):
         return (ingress_address, STANDARD_API_PORT)
     else:
         return (hookenv.unit_public_ip(), STANDARD_API_PORT)
+
+
+def query_cephfs_enabled(ceph_ep):
+    if not is_flag_set('kubernetes-master.ceph-cli.installed'):
+        check_call(['apt', 'install', '-y', 'ceph-common'])
```
Cynerva: `apt update` first?

Cynerva: I bring this up because of all the times where `apt install <foo>` has failed for me because the local manifests are referencing a package that no longer exists.

Do charms regularly run `apt update`? If they do, then there's probably no need to call it here. If they don't, then we probably should call it here to be safe and avoid an endless looping failure.

johnsca: layer-basic includes an optimistic-with-fallback helper, so it would probably be worth using that: https://github.com/juju-solutions/layer-basic/blob/master/lib/charms/layer/basic.py#L218-L251

Cynerva:
Thanks, LGTM :+1:

PR #77, opened 2020-02-25
Add default-cni config

Cynerva:
Partial fix for https://bugs.launchpad.net/charm-kubernetes-master/+bug/1861709

Please be careful merging this, as it's mutually dependent on the work of other PRs in that issue. They will need to be merged together.

This adds the default-cni config to kubernetes-master. The default-cni value is used to select which of multiple cni configs from the cni relation to use. It is also sent via kube-control to kubernetes-worker.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -929,7 +929,8 @@ def start_master():
 
     # kube-proxy
     cni = endpoint_from_flag('cni.available')
-    cluster_cidr = cni.get_config()['cidr']
+    default_cni = hookenv.config('default-cni')
+    cluster_cidr = cni.get_config(default=default_cni)['cidr']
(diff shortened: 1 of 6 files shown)


reactive/kubernetes_master.py
```diff
@@ -2858,3 +2859,15 @@ def api_server_stopped():
     aws_iam = endpoint_from_flag('endpoint.aws-iam.available')
     if aws_iam:
         aws_iam.set_api_server_status(False)
+
+
+@when('kube-control.connected')
+def send_default_cni():
```
kwmonroe: docstring?

Cynerva: added

Cynerva:
In the spirit of similar requests on other PRs, I've added py3.6 and py3.7 to `.travis.yaml` and `tox.ini` here as well.

PR #80, opened 2020-03-05
Send fsName to cdk-addons for CephFS

johnsca:
The name of the CephFS file system created by the ceph-fs charm is not hard-coded. Instead, it is set to the application name that the ceph-fs charm was deployed with, which we cannot determine via Juju because we have no direct relation to ceph-fs. So we have to query ceph to find the correct value to use for fsName.

### DIFF
--- a/actions/create-rbd-pv
+++ b/actions/create-rbd-pv
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 from charms.reactive import is_state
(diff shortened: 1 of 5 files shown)


kwmonroe:
We need this in support of https://bugs.launchpad.net/cdk-addons/+bug/1862236, which updates the cephfs version in cdk-addons.

PR #81, opened 2020-03-11
Write and pass in kubeconfig for cdk-addons

johnsca:
It seems that cdk-addons has always expected a kubeconfig snap config to be passed in with a path to the kubectl config file to use, but k8s-master has never actually created and set that. It happened to work by accident up until now, since kubectl would try localhost:8080 which happened to work. It seems the latest kubectl getting built into cdk-addons no longer falls back to localhost:8080 (see [k8s PR#86173](https://github.com/kubernetes/kubernetes/pull/86173)), so this uncovered the fact that we were not giving an explicit config.


johnsca:
@tvansteenburgh The ones that it would make sense for it to re-use are the ones at either `/root/.kube/config` or `/home/ubuntu/config` but both of those are not visible from within the cdk-addons snap.  The controller-manager and kube-proxy configs are both visible within the snap, but they have specific users and permissions for those services.

PR #87, opened 2020-04-03
Add support for series upgrade

johnsca:
Pause (disable) the services, set an appropriate status.

Part of https://bugs.launchpad.net/charm-kubernetes-master/+bug/1869944

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1167,6 +1167,7 @@ def reconfigure_cdk_addons():
 @when('kubernetes-master.components.started',
       'leadership.is_leader',
       'leadership.set.cluster_tag')
+@when_not('upgrade.series.in-progress')
 def configure_cdk_addons():
     ''' Configure CDK addons '''
     remove_state('cdk-addons.reconfigure')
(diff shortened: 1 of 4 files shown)


Cynerva:
I think it would be good to stop the [configure_cdk_addons](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/a0df3914d38c06fcc06fd23ddb03f6b0cc210871/reactive/kubernetes_master.py#L1170) handler from running during the series upgrade. That handler is going to add a substantial delay to the hook, as cdk-addons times out trying to reach the Kubernetes API while kube-apiserver is no longer up.

The obvious answer is to add `@when_not('upgrade.series.in-progress')` to the configure_cdk_addons handler. I'd be good with that.

Another option is to call `remove_state('kubernetes-master.components.started')` in the pre-series-upgrade hook. That will stop configure_cdk_addons from running until after start_master has run. And start_master won't run because it already has `@when_not('upgrade.series.in-progress')`. The advantage to this approach is that it will not only stop configure_cdk_addons, but also a few other handlers that may try to reach out to the Kubernetes API as well.

Cynerva:
LGTM

PR #89, opened 2020-04-12
Explicitly use bash

damoxc:
Since `[[` is a bash feature and not part of plain sh it should be used instead in the shebang to prevent errors when the wrong shell is configured as default.


Cynerva:
Is there a launchpad issue for this?

PR #91, opened 2020-04-24
Use openstack-integrator to configure k8s-keystone-auth 

lolwww:
Use credentials obtained from openstack-integrator to also configure k8s-keystone-auth which would allow a Kubernetes deployment to avoid using a CMR relation to a controller of the underlying cloud.
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1834164

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1705,7 +1705,8 @@ def build_kubeconfig():
                 tls_ca_filename = 'tls-ca.pem'
                 tls_ca_path = os.path.join(os.sep, 'home', 'ubuntu',
                                            tls_ca_filename)
-                host.write_file(tls_ca_path, base64.b64decode(openstack.endpoint_tls_ca))
+                host.write_file(tls_ca_path,
+                                base64.b64decode(openstack.endpoint_tls_ca))
             context = {'protocol': auth_url_parsed.scheme,
                        'address': auth_url_parsed.hostname,
                        'port': auth_url_parsed.port,
@@ -2498,7 +2499,8 @@ def generate_keystone_configmap():
 
 
 @when('leadership.is_leader', 'kubernetes-master.keystone-policy-handled')
-@when_not_all('keystone-credentials.available.auth', 'endpoint.openstack.ready')
+@when_not_all('keystone-credentials.available.auth',
+              'endpoint.openstack.ready')
 def remove_keystone():
     clear_flag('kubernetes-master.apiserver.configured')
     if not os.path.exists(keystone_policy_path):
@@ -2529,7 +2531,8 @@ def keystone_kick_apiserver():
 
 
 @when('certificates.ca.available', 'certificates.client.cert.available',
-      'authentication.setup', 'etcd.available', 'leadership.set.keystone-cdk-addons-configured')
+      'authentication.setup', 'etcd.available',
+      'leadership.set.keystone-cdk-addons-configured')
 @when_any('keystone-credentials.available.auth', 'endpoint.openstack.ready')
 def keystone_config():
     # first, we have to have the service set up before we can render this stuff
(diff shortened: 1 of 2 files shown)


lolwww:
Hi Cory. Thank you for review.
You raised a good point, maybe we should not force using the creds available from openstack-integrator by default.
Initially we discussed the planned approach on the team call on 6th of April. 
I was hoping we reached verbal agreement the change would be in kubernetes-master charm and I promised to bring the PR soon. 
(office hrs topic - [FE] k8s-master to use credentials obtained from openstack-integrator to configure k8s-keystone-auth)

I do believe this change belongs more in this charm, as the logic to configure auth is already there, and we are just reusing it. 
However openstack-integrator by definition in readme only provides a set of credentials. 
We haven't planned for it to expose keystone-credentials.
The change also seems more complicated, than the current one we have.

Do you think we can compromise by doing something like this?

> if openstack and hookenv.config('use-integrator-creds'):
...

johnsca:
It seems that there were multiple understandings of the item from the office hours discussion, and the item as written is vague enough to be understood in at least three ways:

* Overloading the existing relation, as this PR does (avoids an extra relation; but the only way to make it optional adds to the already extensive config on the k8s-master charm and also doesn't enable similar support for any other charm)
* Just adding keystone credential config to the k8s master charm, as was Tim's understanding (avoids an extra relation, is optional, and would support an entirely external, non-Juju Keystone; but pollutes the k8s-master charm config even more and also doesn't enable similar support for any other charm)
* Adding support for the `keystone-credentials` interface to the openstack-integrator charm, as was my understanding (avoids any changes to the k8s-master charm at all, is cleanly optional, and would work with any other charm that supports the `keystone-credentials` interface; but does require a second or third (if also using `loadbalancer`) relation to the integrator charm)

After a brief discussion with Tim, he was quickly convinced that his approach wasn't a good idea.  If we really wanted to support an external Keystone, the preferred approach would be a proxy charm, as that would isolate the config to that one charm and then serve as a drop-in replacement for any charm that supports the interface.

Adding the relation to the openstack-integrator is essentially like making the integrator charm into a proxy charm.  It also matches the pattern that charm already uses for the loadbalancer relation which is used to replace the kubeapi-load-balancer with a native LB (other integrators are expected to grow this as well).  Overall, it feels to me like it fits best with the Juju aesthetic, where interfaces have a well-defined semantic meaning and act as connection points for features that the charm provides or requires, and allows charms to be connected together in composable ways.

> I do believe this change belongs more in this charm, as the logic to configure auth is already there, and we are just reusing it.

Having the integrator charm grow support for the existing interface would even more directly reuse the existing auth logic, since no changes at all would be necessary to k8s-master and the integrator would just be copying existing data to the new relation.

> openstack-integrator by definition in readme only provides a set of credentials.

The general idea of the integrator charms is to act as proxy charms for cloud functionality.  The loadbalancer relation is an example of this, but even the `openstack-integration` interface [provides more than just credentials](https://github.com/juju-solutions/interface-openstack-integration/blob/master/docs/requires.md) for the things that don't map well to an existing interface.  (I do somewhat regret overloading that relation as much as it is already; a lot of those values could be handled much more cleanly with a new "workload loadbalancer" interface instead.)

I should also also note that the current relation creates a separate `k8s` user which could conceivably be done by the integrator charm so that it didn't need to use the project admin creds, but that effort might not be worth it.

> The change also seems more complicated, than the current one we have.

I was originally going to say that it seems to me to be the opposite, since k8s charm wouldn't need to have any of these conditionals and the integrator charm would in theory just be copying data into an additional relation (just in a slightly different format), but unfortunately I see that the [`keystone-credentials` interface](https://github.com/openstack/charm-interface-keystone-credentials) doesn't have a provides-side implementation, so that would need to be added.  Over all, I think it's worth it for a cleaner conceptual model, though.

lolwww:
As discissed, done in integrator 
https://github.com/juju-solutions/charm-openstack-integrator/pull/38

johnsca:
Closing in favor of https://github.com/juju-solutions/charm-openstack-integrator/pull/38 and https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/127

PR #93, opened 2020-04-29
support for service-cidr expansion

hemanthnakkina:
kubernetes-master charm does not allow service-cidr to be changed
once k8s is deployed. However k8s by itself supports service-cidr
expansion. This PR adds support for service-cidr expansion in
kubernetes-master charm.

The charm checks whether the new service-cidr is an expansion or not.
If it is not an expansion, a warning message is displayed on the
status line.

For expansion case,
* regenerate API certs
* update kube-apiserver parameters
* restart kube-apiserver apiservice
* delete kubernetes service object
* rollout cdk-addon services

Closes-Bug: #1868685

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -168,6 +168,42 @@ def channel_changed():
     set_upgrade_needed()
 
 
+# Returns True if a is subnet of b
+# This method is copied from cpython as it is available only from
+# python 3.7
+# https://github.com/python/cpython/blob/3.7/Lib/ipaddress.py#L1000
+def is_subnet_of(a, b):
+    try:
+        # Always false if one is v4 and the other is v6.
+        if a._version != b._version:
+            raise TypeError("{} and {} are not of the same version".format(
+                a, b))
+        return (b.network_address <= a.network_address and
+                b.broadcast_address >= a.broadcast_address)
+    except AttributeError:
+        raise TypeError("Unable to test subnet containment "
+                        "between {} and {}".format(a, b))
+
+
+def is_service_cidr_expansion():
+    service_cidr_from_db = db.get('kubernetes-master.service-cidr')
+    service_cidr_from_config = hookenv.config('service-cidr')
+    if not service_cidr_from_db:
+        return False
+
+    # Do not consider as expansion if both old and new service cidr are same
+    if service_cidr_from_db == service_cidr_from_config:
+        return False
+
+    current_service_cidr = ipaddress.ip_network(service_cidr_from_db)
+    new_service_cidr = ipaddress.ip_network(service_cidr_from_config)
+    if not is_subnet_of(current_service_cidr, new_service_cidr):
+        hookenv.log("WARN: New k8s service cidr not superset of old one")
+        return False
+
+    return True
+
+
 def service_cidr():
     ''' Return the charm's service-cidr config '''
     frozen_cidr = db.get('kubernetes-master.service-cidr')
@@ -177,7 +213,10 @@ def service_cidr():
 def freeze_service_cidr():
     ''' Freeze the service CIDR. Once the apiserver has started, we can no
     longer safely change this value. '''
-    db.set('kubernetes-master.service-cidr', service_cidr())
+    frozen_service_cidr = db.get('kubernetes-master.service-cidr')
+    if not frozen_service_cidr or is_service_cidr_expansion():
+        db.set('kubernetes-master.service-cidr', hookenv.config(
+            'service-cidr'))
 
 
 def maybe_install_kube_proxy():
@@ -1594,7 +1633,8 @@ def on_config_allow_privileged_change():
 @when_any('config.changed.api-extra-args',
           'config.changed.audit-policy',
           'config.changed.audit-webhook-config',
-          'config.changed.enable-keystone-authorization')
+          'config.changed.enable-keystone-authorization',
+          'config.changed.service-cidr')
 @when('kubernetes-master.components.started')
 @when('leadership.set.auto_storage_backend')
 @when('etcd.available')
@@ -1801,6 +1841,10 @@ def configure_apiserver():
         # No point in trying to start master services and fail. Just return.
         return
 
+    # Update unit db service-cidr
+    was_service_cidr_expanded = is_service_cidr_expansion()
+    freeze_service_cidr()
+
     api_opts = {}
 
     if is_privileged():
@@ -1973,6 +2017,37 @@ def configure_apiserver():
     configure_kubernetes_service(configure_prefix, 'kube-apiserver',
                                  api_opts, 'api-extra-args')
     service_restart('snap.kube-apiserver.daemon')
+
+    if was_service_cidr_expanded and is_state('leadership.is_leader'):
+        try:
+            hookenv.log("service-cidr expansion: Deleting service kubernetes")
+            kubectl('delete', 'service', 'kubernetes')
+
+            # Restart the cdk-addons
+            # Get deployments/daemonsets/statefulsets
+            hookenv.log("service-cidr expansion: Restart the cdk-addons")
+            output = kubectl(
+                'get', 'daemonset,deployment,statefulset',
+                '-o', 'json',
+                '--all-namespaces',
+                '-l', 'cdk-restart-on-ca-change=true'
+            ).decode('UTF-8')
+            deployments = json.loads(output)['items']
+
+            # Now restart the addons
+            for deployment in deployments:
+                kind = deployment['kind']
+                namespace = deployment['metadata']['namespace']
+                name = deployment['metadata']['name']
+                hookenv.log('Restarting addon: {0} {1} {2}'.format(kind,
+                                                                   namespace,
+                                                                   name))
+                kubectl(
+                    'rollout', 'restart', kind + '/' + name,
+                    '-n', namespace
+                )
+        except Exception:
+            hookenv.log("service-cidr-expansion: k8s services not yet started")
 
     set_flag('kubernetes-master.apiserver.configured')
 
--- a/tests/test_kubernetes_master.py
+++ b/tests/test_kubernetes_master.py
@@ -1,8 +1,9 @@
 import pytest
 from unittest import mock
 from reactive import kubernetes_master
+from charms.layer.kubernetes_common import get_version, kubectl
 from charms.reactive import endpoint_from_flag, remove_state
-from charmhelpers.core import hookenv
+from charmhelpers.core import hookenv, unitdata
 
 
 def patch_fixture(patch_target):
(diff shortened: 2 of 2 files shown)


reactive/kubernetes_master.py
```diff
@@ -1801,6 +1844,10 @@ def configure_apiserver():
         # No point in trying to start master services and fail. Just return.
         return
 
+    if not is_service_cidr_expansion():
+        hookenv.log("WARN: service cidr changed but not an expansion")
+        return
```
johnsca, line 1849: This will block all subsequent changes to the apiserver config, rather than simply ignoring the non-expansion CIDR change.

reactive/kubernetes_master.py
```diff
@@ -168,16 +168,45 @@ def channel_changed():
...
+            return False
+
+    return True
+
+
 def service_cidr():
     ''' Return the charm's service-cidr config '''
-    frozen_cidr = db.get('kubernetes-master.service-cidr')
-    return frozen_cidr or hookenv.config('service-cidr')
+    return hookenv.config('service-cidr')
```
johnsca, line 202: This change seems unsafe and unnecessary.  I think this should be reverted and `freeze_service_cidr` changed as mentioned below.

reactive/kubernetes_master.py
```diff
@@ -168,16 +168,45 @@ def channel_changed():
...
 
 
 def freeze_service_cidr():
     ''' Freeze the service CIDR. Once the apiserver has started, we can no
     longer safely change this value. '''
-    db.set('kubernetes-master.service-cidr', service_cidr())
+    if is_service_cidr_expansion():
+        db.set('kubernetes-master.service-cidr', service_cidr())
```
johnsca, line 209: The call to `service_cidr()` here should be changed to `hookenv.config('service-cidr')` rather than changing it in `service_cidr()` above.

johnsca:
hemanthnakkina, thanks for the improvement.  I requested some changes above because of a corner case where if the `service-cidr` config was changed either along with or just prior to any other apiservice relevant configs, the latter would get ignored along with the `service-cidr` change.  Of course, the status message would indicate the issue and once the `service-cidr` was corrected, the other settings would be applied, but it seems safer overall to maintain the current behavior of ignoring only the `service-cidr` change when it is not an expansion.

johnsca:
Oh, it would also be best if one or more tests could be added to [the existing unit tests](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/tests/test_kubernetes_master.py) to cover this, since there's a fair bit of logic in it.

hemanthnakkina:
johnsca, thanks for your comments. I have submitted a new patch as per your comments.
Yet to handle unit tests and integration tests. Will request for re-review once the units tests and integration tests are in place.

johnsca:
+1 looks good and I look forward to the tests.  Thank again!

hemanthnakkina:
@johnsca 
- Added unit tests in this patch
- Integration tests are added here - https://github.com/charmed-kubernetes/jenkins/pull/594

Can you let me know how to trigger the tests for this PR with the modified integration test PR ( any tag need to be added in commit message?)

reactive/kubernetes_master.py
```diff
@@ -1974,6 +2020,38 @@ def configure_apiserver():
                                  api_opts, 'api-extra-args')
     service_restart('snap.kube-apiserver.daemon')
 
+    is_leader = is_state('leadership.is_leader')
+    if is_leader and is_service_cidr_expansion():
```
Cynerva, line 2024: Can you update this if condition (or update `is_service_cidr_expansion`) to also check that the service-cidr has changed?

As it is now, it looks like if the service-cidr hasn't changed, then `is_subnet_of(current_service_cidr, new_service_cidr)` will return True, therefore `is_service_cidr_expansion()` will return True, therefore this condition will be True on the leader.

As a result, every time the charm reconfigures kube-apiserver, even if service-cidr hasn't changed, it will delete the kubernetes service and do a rolling restart of addons. That's a bit too disruptive for us to be doing it when we don't need to.

reactive/kubernetes_master.py
```diff
@@ -1754,6 +1786,17 @@ def get_deprecated_dns_ip():
     return ip.exploded
 
 
+def get_k8s_cluster_ip():
```
Cynerva, line 1789: This function is unused, can you remove it?

reactive/kubernetes_master.py
johnsca: Ah, good catch.  However, checking in either place will not work without additional changes since the new CIDR is already frozen by the time we get here.

```suggestion
    if was_service_cidr_expanded and is_state('leadership.is_leader'):
```

reactive/kubernetes_master.py
```diff
@@ -168,6 +168,35 @@ def channel_changed():
...
+        raise TypeError("Unable to test subnet containment "
+                        "between {} and {}".format(a, b))
+
+
+def is_service_cidr_expansion():
+    service_cidr_from_db = db.get('kubernetes-master.service-cidr')
+    if service_cidr_from_db:
+        current_service_cidr = ipaddress.ip_network(service_cidr_from_db)
+        new_service_cidr = ipaddress.ip_network(hookenv.config('service-cidr'))
+        if not is_subnet_of(current_service_cidr, new_service_cidr):
```
johnsca, line 193:
```suggestion
        if new_service_cidr == current_service_cidr:
            return False
        if not is_subnet_of(current_service_cidr, new_service_cidr):
```

reactive/kubernetes_master.py
```diff
@@ -177,7 +206,9 @@ def service_cidr():
 def freeze_service_cidr():
     ''' Freeze the service CIDR. Once the apiserver has started, we can no
     longer safely change this value. '''
-    db.set('kubernetes-master.service-cidr', service_cidr())
+    if is_service_cidr_expansion():
+        db.set('kubernetes-master.service-cidr', hookenv.config(
+            'service-cidr'))
```
johnsca, line 211:
```suggestion
    db.set('kubernetes-master.service-cidr', hookenv.config('service-cidr'))
```

Suggest moving the check outside the function to allow for caching the expansion check result, see below.

reactive/kubernetes_master.py
```diff
@@ -1801,6 +1844,9 @@ def configure_apiserver():
         # No point in trying to start master services and fail. Just return.
         return
 
+    # Update unit db service-cidr
+    freeze_service_cidr()
```
johnsca, line 1848:
```suggestion
    was_service_cidr_expanded = is_service_cidr_expansion()
    if was_service_cidr_expanded:
        # Update unit db service-cidr
        freeze_service_cidr()
```

This has to be cached or we won't be able to tell below whether there was an actual change.

reactive/kubernetes_master.py
Cynerva: This change helps, but `is_service_cidr_expansion()` would still return True when `kubernetes-master.service-cidr` hasn't been set yet. I don't think that's what we want - we don't want to be deleting the kubernetes service when the cluster is first deployed.

reactive/kubernetes_master.py
Cynerva: This change will make it necessary to edit or remove the call to freeze_service_cidr here: https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/d89d04d3b0a2c12fe7f676a5bdd130334a442f27/reactive/kubernetes_master.py#L965

johnsca: Good point.  And it's not actually necessary to move the check outside of the freeze function, so it's probably better to skip this change as well as adding the `if` around the other call.  I'll update / remove the suggested changes to avoid confusion.

reactive/kubernetes_master.py
johnsca: Good point.  We actually want different behaviors for `freeze_service_cidr()` vs the rolling restart, since the former should be done on the initial deploy but the latter shouldn't.

reactive/kubernetes_master.py
johnsca:
```suggestion
    frozen_service_cidr = db.get('kubernetes-master.service-cidr')
    if not frozen_service_cidr or is_service_cidr_expansion():
        db.set('kubernetes-master.service-cidr', hookenv.config(
            'service-cidr'))
```

reactive/kubernetes_master.py
johnsca: Per @Cynerva's comment elsewhere, this should instead be:

```suggestion
    was_service_cidr_expanded = is_service_cidr_expansion()
    freeze_service_cidr()
```

reactive/kubernetes_master.py
```diff
@@ -168,6 +168,35 @@ def channel_changed():
...
+def is_service_cidr_expansion():
+    service_cidr_from_db = db.get('kubernetes-master.service-cidr')
+    if service_cidr_from_db:
+        current_service_cidr = ipaddress.ip_network(service_cidr_from_db)
+        new_service_cidr = ipaddress.ip_network(hookenv.config('service-cidr'))
+        if not is_subnet_of(current_service_cidr, new_service_cidr):
+            hookenv.log("WARN: New k8s service cidr not superset of old one")
+            return False
+
+    return True
```
johnsca, line 197:
```suggestion
def is_service_cidr_expansion():
    service_cidr_from_db = db.get('kubernetes-master.service-cidr')
    if not service_cidr_from_db:
        return False
    current_service_cidr = ipaddress.ip_network(service_cidr_from_db)
    new_service_cidr = ipaddress.ip_network(hookenv.config('service-cidr'))
    if not is_subnet_of(current_service_cidr, new_service_cidr):
        hookenv.log("WARN: New k8s service cidr not superset of old one")
        return False

    return True
```

Per @Cynerva's comment, I think the non-frozen-CIDR case result should be inverted and handled explicitly in `freeze_service_cidr()`

reactive/kubernetes_master.py
hemanthnakkina: Modified as per your suggestion

reactive/kubernetes_master.py
hemanthnakkina: Modified as per your suggestion. New check added to return False if old and new service cidr is same.

hemanthnakkina:
@Cynerva @johnsca thanks for your time to review the patch and for the suggestions. Made all the changes, please do re-review.

Cynerva:
LGTM

johnsca:
This will be in the next edge bundle when it gets built, which should be `cs:charmed-kubernetes-449` or greater.

johnsca:
The Jenkins test for this has been enabled and will be visible on the next set of runs (probably over night, as 5/13).  You can check the tests here: http://jenkaas.s3-website-us-east-1.amazonaws.com/

PR #99, opened 2020-05-28
Support IPv6 (only or dual-stack)

johnsca:
Part of [lp:1881177](https://bugs.launchpad.net/charm-etcd/+bug/1881177)


johnsca:
This hasn't been tested for IPv6-aware pods trying to talk to the K8s API server.  I think we'll need to address both [lp:1857173](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1857173) and [lp:1809379](https://bugs.launchpad.net/charm-etcd/+bug/1809379) to handle that case.

Cynerva:
> This hasn't been tested for IPv6-aware pods trying to talk to the K8s API server. I think we'll need to address both lp:1857173 and lp:1809379 to handle that case.

Those issues don't look relevant to me -- pods do not talk to kubeapi-load-balancer or etcd. I think it should be fine?

lib/charms/layer/kubernetes_master.py
```diff
@@ -144,3 +149,87 @@ def deprecate_auth_file(auth_file):
...
+    return kubernetes_common.get_service_ip('kube-dns',
+                                            namespace='kube-system')
+
+
+def get_deprecated_dns_ip():
+    '''We previously hardcoded the dns ip. This function returns the old
+    hardcoded value for use with older versions of cdk_addons.'''
+    network = get_preferred_service_network(service_cidr())
+    ip = next(islice(network.hosts(), 9, None))
+    return ip.exploded
```
Cynerva, line 229: Let's just get rid of `get_deprecated_dns_ip`. It is only used here:

https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/472f0bf992d994bc444dbfa7ceabe337cfce4db3/reactive/kubernetes_master.py#L1280

The `dns-ip` option to cdk-addons has been removed since cdk-addons 1.9. We don't need to specify it any more.

johnsca:
@Cynerva That comment came from the issue that came up in the original discovery which used IPv6-only for the API server endpoint (and also the service in front of it that the pods use).  However, on further reflection I think that was actually an issue with the etcd endpoint being IPv6 rather than the K8s API endpoint, so I think you're correct that those are not actually an issue with the current dual-stack approach.

reactive/kubernetes_master.py
```diff
@@ -1036,21 +987,23 @@ def start_master():
...
     service_restart('snap.kube-proxy.daemon')
 
     set_state('kubernetes-master.components.started')
     hookenv.open_port(6443)
 
 
+@when('config.changed.proxy-extra-args')
+def proxy_args_changed():
+    clear_flag('kubernetes-master.components.started')
+    clear_flag('config.changed.proxy-extra-args')
```
Cynerva, line 1004: This change fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1841112. I've linked this PR to the issue and assigned you accordingly.

reactive/kubernetes_master.py
```diff
@@ -1202,35 +1156,36 @@ def push_service_data():
...
+    bind_ips = kubernetes_common.get_bind_addrs(
+        ipv4=kubernetes_common.is_ipv4(cluster_cidr),
+        ipv6=kubernetes_common.is_ipv6(cluster_cidr),
+    )
 
     domain = hookenv.config('dns_domain')
     # Create SANs that the tls layer will add to the server cert.
     sans = [
-        hookenv.unit_public_ip(),
-        ingress_ip,
```
Cynerva, line 1224: Why is it safe to remove ingress_ip? I would expect it to still be needed - it's the IP that clients of the kube-api-endpoint relation use to connect to the API.

johnsca: It should be covered by `get_bind_addrs()` but I'll add it back explicitly just in case it isn't.

Cynerva:
LGTM

PR #101, opened 2020-06-08
Fix service-cidr expansion race condition

johnsca:
The logic for handling service-cidr expansion doesn't gracefully handle the case where the API server doesn't come up quickly enough. There's also some question as to whether the API server might need to be restarted after the service is deleted, so check for and handle that.

Fixes [lp:1879785](https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1879785)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -822,6 +822,11 @@ def set_final_status():
                                       'to secure secrets')
         return
 
+    if is_state('kubernetes-master.had-service-cidr-expanded'):
+        hookenv.status_set('waiting',
+                           'Waiting to retry updates for service-cidr expansion')
+        return
+
     if is_state('kubernetes-master.components.started'):
         # All services should be up and running at this point. Double-check...
         failing_services = master_services_down()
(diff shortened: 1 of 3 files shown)


reactive/kubernetes_master.py
```diff
@@ -2063,37 +2068,77 @@ def configure_apiserver():
...
+            namespace = deployment['metadata']['namespace']
+            name = deployment['metadata']['name']
+            hookenv.log('Restarting addon: {0} {1} {2}'.format(kind,
+                                                               namespace,
+                                                               name))
+            kubectl(
+                'rollout', 'restart', kind + '/' + name,
+                '-n', namespace
+            )
+    except CalledProcessError:
```
Cynerva, line 2138: `hookenv.log(traceback.format_exc())` please

Cynerva: Per IRC, the kubectl calls are already logged here: https://github.com/charmed-kubernetes/layer-kubernetes-common/blob/be49e03a77e9abee648aba39e707ea39d8982427/lib/charms/layer/kubernetes_common.py#L220

So that should be good enough to figure out what's going on when this fails.

Cynerva:
LGTM

PR #102, opened 2020-06-16
Fix vSphere host UUID discovery

sombrafam:
When upgrading the vSphere hardware from 5.5 -> 6.5 no more PVCs
can be created. This happens because of a bug on the SEABIOS virtual
hardware that swap the endianness of the content of '/sys/class/dmi/id/product_uuid'
on the new virtual hardware versions and now it differs from
'/sys/class/dmi/id/product_serial'.
So, the content in product_uuid is: F5132842-20D4-6171-5DB0-7B96440CCF54
While the product_serial (actually correct and understood by vSphere API)
is:422813f5-d420-7161-5db0-7b96440ccf54

 So, we need to use the product_serial instead of product_uuid

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -86,6 +86,7 @@ from charms.layer.kubernetes_common import server_key_path
 from charms.layer.kubernetes_common import client_crt_path
 from charms.layer.kubernetes_common import client_key_path
 from charms.layer.kubernetes_common import kubectl, kubectl_manifest, kubectl_success
+from charms.layer.kubernetes_common import _get_vmware_uuid
 
 from charms.layer.nagios import install_nagios_plugin_from_file
 from charms.layer.nagios import remove_nagios_plugin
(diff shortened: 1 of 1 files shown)


sombrafam:
> Would it be worth checking `product_uuid` in case `product_serial` fails? If `_serial` is available on both and conforms to this check, great; if not, it'd be nice to handle both 5.x and 6.x.

The product_serial and product_uuid are only and always present for vSphere VMs. So, checking the 2 won't add any
resiliency. Also, if we still want to rely on product_uuid, we would need to fix the endianness which is not a good idea given that 
VMWare can fix the bug and our logic would break. By using only product_serial, both 5.x and 6.x works.

kwmonroe (changes requested):
Code LGTM, but given this affects the master and [worker](https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/64), and we have 2 PRs for the same func, would you mind refactoring to drop `_get_vmware_uuid` into https://github.com/charmed-kubernetes/layer-kubernetes-common/blob/master/lib/charms/layer/kubernetes_common.py?

Unless I'm missing something, it seems better to me if both -master and -worker utilize the same function.

kwmonroe (approved):
LGTM; thanks for the refactor

PR #104, opened 2020-07-14
Add CIS benchmark requirements to k8s master

joedborg:
This results in 

```
[FAIL] 1.2.3 Ensure that the --token-auth-file parameter is not set (Scored)
```

For the API server tests, I believe these will be addressed in the move to webhook auth - please advise if this is incorrect.  Will address the scheduler and controller in other PRs.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1964,11 +1964,11 @@ def configure_apiserver():
 
     admission_plugins = [
         'PersistentVolumeLabel',
+        'PodSecurityPolicy',
+        'NodeRestriction'
     ]
 
     auth_mode = hookenv.config('authorization-mode')
-    if 'Node' in auth_mode:
-        admission_plugins.append('NodeRestriction')
 
     ks = endpoint_from_flag('keystone-credentials.available.auth')
     aws = endpoint_from_flag('endpoint.aws-iam.ready')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1007,8 +1007,6 @@ def add_systemd_file_watcher():
           'tls_client.certs.changed',
           'tls_client.ca.written',
           'upgrade.series.in-progress')
-
-
 def start_master():
     '''Run the Kubernetes master components.'''
     hookenv.status_set('maintenance',
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -2187,6 +2187,9 @@ def configure_controller_manager():
     controller_opts['tls-cert-file'] = str(server_crt_path)
     controller_opts['tls-private-key-file'] = str(server_key_path)
     controller_opts['cluster-name'] = leader_get('cluster_tag')
+    controller_opts['terminated-pod-gc-threshold'] = '12500'
+    controller_opts['profiling'] = 'false'
+    controller_opts['feature-gates'] = 'RotateKubeletServerCertificate=true'
 
     cm_cloud_config_path = cloud_config_path('kube-controller-manager')
     if is_state('endpoint.aws.ready'):
@@ -2214,6 +2217,7 @@ def configure_scheduler():
     scheduler_opts['v'] = '2'
     scheduler_opts['logtostderr'] = 'true'
     scheduler_opts['master'] = 'http://127.0.0.1:8080'
+    scheduler_opts['profiling'] = 'false'
 
     configure_kubernetes_service(configure_prefix, 'kube-scheduler',
                                  scheduler_opts, 'scheduler-extra-args')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1582,7 +1582,11 @@ def switch_auth_mode(forced=False):
 def create_pod_security_policy_resources():
     pod_security_policy_path = '/root/cdk/pod-security-policy.yaml'
 
-    render('rbac-pod-security-policy.yaml')
+    render(
+        'rbac-pod-security-policy.yaml',
+        pod_security_policy_path,
+        {}
+    )
 
     hookenv.log('Creating pod security policy resources.')
     if kubectl_manifest('apply', pod_security_policy_path):
@@ -1800,8 +1804,6 @@ def build_kubeconfig():
 
     public_address, public_port = kubernetes_master.get_api_endpoint()
     public_server = 'https://{0}:{1}'.format(public_address, public_port)
-    local_address = get_ingress_address('kube-api-endpoint')
-    local_server = 'https://{0}:{1}'.format(local_address, 6443)
     ca_exists = ca_crt_path.exists()
     client_pass = get_token('admin')
     # Do we have everything we need?
@@ -1848,21 +1850,21 @@ def build_kubeconfig():
 
         # make a copy in a location shared by kubernetes-worker
         # and kubernete-master
-        create_kubeconfig(kubeclientconfig_path, local_server, ca_crt_path,
+        create_kubeconfig(kubeclientconfig_path, public_server, ca_crt_path,
                           user='admin', token=client_pass)
 
         # make a copy for cdk-addons to use
-        create_kubeconfig(cdk_addons_kubectl_config_path, local_server,
+        create_kubeconfig(cdk_addons_kubectl_config_path, public_server,
                           ca_crt_path, user='admin', token=client_pass)
 
         # make a kubeconfig for kube-proxy
         proxy_token = get_token('system:kube-proxy')
-        create_kubeconfig(kubeproxyconfig_path, local_server, ca_crt_path,
+        create_kubeconfig(kubeproxyconfig_path, public_server, ca_crt_path,
                           token=proxy_token, user='kube-proxy')
 
         controller_manager_token = get_token('system:kube-controller-manager')
         create_kubeconfig(kubecontrollermanagerconfig_path,
-                          local_server, ca_crt_path,
+                          public_server, ca_crt_path,
                           token=controller_manager_token,
                           user='kube-controller-manager')
 
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1157,7 +1157,6 @@ def create_tokens_and_sign_auth_requests():
     scheduler_token = get_token('system:kube-scheduler')
     if not scheduler_token:
         setup_tokens(None, 'system:kube-scheduler', 'system:kube-scheduler')
-        scheduler_token = get_token('system:kube-scheduler')
 
     client_token = get_token('admin')
     if not client_token:
@@ -1183,7 +1182,7 @@ def create_tokens_and_sign_auth_requests():
             kubelet_token = get_token(username)
         kube_control.sign_auth_request(request[0], username,
                                        kubelet_token, proxy_token,
-                                       scheduler_token, client_token)
+                                       client_token)
 
 
 @when('kube-api-endpoint.available')
@@ -2421,10 +2420,13 @@ def poke_network_unavailable():
     discussion about refactoring the affected code but nothing has happened
     in a while.
     """
-    public_address, public_port = kubernetes_master.get_api_endpoint()
-    public_server = 'https://{0}:{1}'.format(public_address, public_port)
+    local_address = get_ingress_address('kube-api-endpoint')
+    local_server = 'https://{0}:{1}'.format(local_address, 6443)
 
     cmd = ['kubectl', 'get', 'nodes', '-o', 'json']
+
+    client_token = get_token('admin')
+    http_header = ('Authorization', 'Bearer {}'.format(client_token))
 
     try:
         output = check_output(cmd).decode('utf-8')
@@ -2439,8 +2441,9 @@ def poke_network_unavailable():
 
     for node in nodes:
         node_name = node['metadata']['name']
-        url = '{}/api/v1/nodes/{}/status'.format(public_server, node_name)
-        with urlopen(url) as response:
+        req = Request('{}/api/v1/nodes/{}/status'.format(local_server, node_name))
+        req.add_header(*http_header)
+        with urlopen(req) as response:
             code = response.getcode()
             body = response.read().decode('utf8')
         if code != 200:
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1944,6 +1944,8 @@ def configure_apiserver():
     api_opts['kubelet-preferred-address-types'] = \
         'InternalIP,Hostname,InternalDNS,ExternalDNS,ExternalIP'
     api_opts['advertise-address'] = get_ingress_address('kube-control')
+    api_opts['encryption-provider-config'] = \
+        str(encryption_config_path())
 
     etcd_dir = '/root/cdk/etcd'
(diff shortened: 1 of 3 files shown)


reactive/kubernetes_master.py
```diff
@@ -1011,6 +1011,24 @@ def start_master():
     '''Run the Kubernetes master components.'''
     hookenv.status_set('maintenance',
                        'Configuring the Kubernetes master services.')
+
+    if not is_state('kubernetes-master.vault-kv.pending'):
```
kwmonroe, line 1015: We're doing this to ensure we don't stomp on an encryption config file set up by the vault relation, right?  If so, I think it would be better to guard with `not is_state('kubernetes-master.secure-storage.created')`, or perhaps both flags -- not `vault-kv.pending` and not `secure-storage.created`.

I'm nervous that we'll enter `start_master` when vault has been configured, so it's no longer pending.  At that point, we'd overwrite whatever vault config we had with this.

joedborg:
Added scheduler and controller work too

Cynerva:
> ```
> [FAIL] 1.2.7 Ensure that the --authorization-mode argument is not set to AlwaysAllow (Scored)
> [FAIL] 1.2.8 Ensure that the --authorization-mode argument includes Node (Scored)
> [FAIL] 1.2.9 Ensure that the --authorization-mode argument includes RBAC (Scored)
> ```
> For the API server tests, I believe these will be addressed in the move to webhook auth - please advise if this is incorrect.

As I understand it, the webhook auth thing will not address the --authorization-mode config. When we talk about "webhook auth" we're talking about authentication, not authorization.

reactive/kubernetes_master.py
```diff
@@ -1944,11 +1965,11 @@ def configure_apiserver():
 
     admission_plugins = [
         'PersistentVolumeLabel',
+        'PodSecurityPolicy',
```
Cynerva, line 1968: Yikes, PodSecurityPolicy is a can of worms. Please check out the [upstream docs on Pod Security Policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) and in particular, note this warning:

> PodSecurityPolicies are enforced by enabling the admission controller, but doing so without authorizing any policies will prevent any pods from being created in the cluster.

I believe if we are enabling PodSecurityPolicy, then the charm will also need to create one or more PodSecurityPolicies, along with Roles/ClusterRoles and RoleBindings/ClusterRoleBindings to authorize users to use those policies. It's gonna get messy.

reactive/kubernetes_master.py
```diff
@@ -2059,7 +2081,10 @@ def configure_apiserver():
     else:
         remove_if_exists(audit_webhook_config_path)
 
-    if is_flag_set('kubernetes-master.secure-storage.created'):
+    if kube_version > (1, 8):
```
Cynerva, line 2084: I wouldn't even bother with this check personally. Charmed Kubernetes 1.19 does not need to support anything earlier than Kubernetes 1.17.

reactive/kubernetes_master.py
joedborg: Yeah, seems to be.  When I set `RBAC,Node`, everything stops working because (I'm guessing) these aren't set.  It's a requirement of the CIS benchmark though.

reactive/kubernetes_master.py
```diff
@@ -2162,6 +2187,9 @@ def configure_controller_manager():
     controller_opts['tls-cert-file'] = str(server_crt_path)
     controller_opts['tls-private-key-file'] = str(server_key_path)
     controller_opts['cluster-name'] = leader_get('cluster_tag')
+    controller_opts['terminated-pod-gc-threshold'] = '12500'
+    controller_opts['profiling'] = 'false'
+    controller_opts['feature-gates'] = 'RotateKubeletServerCertificate=true'
```
Cynerva, line 2192: Not sure what the impact of enabling `RotateKubeletServerCertificate` will be for us, given that we pass our own certificate into Kubelet rather than allowing it to generate its own certificate through the k8s Certificates API.

Hopefully this is a no-op, but we should keep our eye out for cert issues relating to Kubelet. In particular, we should make sure the `kubectl logs` and `kubectl exec` commands still work.

reactive/kubernetes_master.py
joedborg: Ah, no, I misspoke, it just took a while to settle after changing to `RBAC,Node`.  It seems to work without these.

reactive/kubernetes_master.py
joedborg: I'm not sure either.  Both `logs` and `exec` work though.  Is it a worry they'll work until rotated?

Cynerva:
> So, would the only way to fix this be to set this as `RBAC,Node` by default?

I believe so.

joedborg:
```
[FAIL] 1.2.3 Ensure that the --token-auth-file parameter is not set (Scored)
1 checks FAIL
```

When `RBAC,Node` is enabled, this is the only failure left.

reactive/kubernetes_master.py
Cynerva: Are you sure? If I enable PodSecurityPolicy on a 1.18 cluster with:

```
juju config kubernetes-master api-extra-args=enable-admission-plugins=PersistentVolumeLabel,PodSecurityPolicy,NodeRestriction
```

At first, everything looks fine:

```
$ kubectl get po --all-namespaces
NAMESPACE                         NAME                                                      READY   STATUS    RESTARTS   AGE
ingress-nginx-kubernetes-worker   default-http-backend-kubernetes-worker-7f969db96f-jpxjh   1/1     Running   0          9m38s
ingress-nginx-kubernetes-worker   nginx-ingress-controller-kubernetes-worker-j9blx          1/1     Running   0          9m30s
kube-system                       coredns-6b59b8bd9f-2gnh5                                  1/1     Running   0          9m54s
kube-system                       kube-state-metrics-69f474f8cb-2dbck                       1/1     Running   0          9m54s
kube-system                       metrics-server-v0.3.6-74c87686d-zwm89                     2/2     Running   0          9m54s
kubernetes-dashboard              dashboard-metrics-scraper-78d587b6b-77g4p                 1/1     Running   0          9m54s
kubernetes-dashboard              kubernetes-dashboard-7b76878768-zjnj4                     1/1     Running   0          9m54s
```

But if I delete one of the pods, it never comes back:

```
$ kubectl delete po -n kubernetes-dashboard kubernetes-dashboard-7b76878768-zjnj4
pod "kubernetes-dashboard-7b76878768-zjnj4" deleted

$ kubectl get po --all-namespaces
NAMESPACE                         NAME                                                      READY   STATUS    RESTARTS   AGE
ingress-nginx-kubernetes-worker   default-http-backend-kubernetes-worker-7f969db96f-jpxjh   1/1     Running   0          10m
ingress-nginx-kubernetes-worker   nginx-ingress-controller-kubernetes-worker-j9blx          1/1     Running   0          10m
kube-system                       coredns-6b59b8bd9f-2gnh5                                  1/1     Running   0          11m
kube-system                       kube-state-metrics-69f474f8cb-2dbck                       1/1     Running   0          11m
kube-system                       metrics-server-v0.3.6-74c87686d-zwm89                     2/2     Running   0          11m
kubernetes-dashboard              dashboard-metrics-scraper-78d587b6b-77g4p                 1/1     Running   0          11m
```

and when I check the kube-controller-manager logs, I see it trying and failing to create the pod:

```
$ journalctl -o cat -u snap.kube-controller-manager.daemon | tail -n 8
I0715 16:09:33.539660    2619 replica_set.go:562] Too few replicas for ReplicaSet kubernetes-dashboard/kubernetes-dashboard-7b76878768, need 1, creating 1
I0715 16:09:33.544646    2619 replica_set.go:587] Slow-start failure. Skipping creation of 1 pods, decrementing expectations for ReplicaSet kubernetes-dashboard/kubernetes-dashboard-7b76878768
E0715 16:09:33.544731    2619 replica_set.go:535] sync "kubernetes-dashboard/kubernetes-dashboard-7b76878768" failed with pods "kubernetes-dashboard-7b76878768-" is forbidden: no providers available to validate pod request
I0715 16:09:33.544773    2619 event.go:278] Event(v1.ObjectReference{Kind:"ReplicaSet", Namespace:"kubernetes-dashboard", Name:"kubernetes-dashboard-7b76878768", UID:"663e530b-8726-4619-88c8-d8374a6226f2", APIVersion:"apps/v1", ResourceVersion:"1878", FieldPath:""}): type: 'Warning' reason: 'FailedCreate' Error creating: pods "kubernetes-dashboard-7b76878768-" is forbidden: no providers available to validate pod request
I0715 16:11:46.857978    2619 replica_set.go:562] Too few replicas for ReplicaSet kubernetes-dashboard/kubernetes-dashboard-7b76878768, need 1, creating 1
I0715 16:11:46.862448    2619 replica_set.go:587] Slow-start failure. Skipping creation of 1 pods, decrementing expectations for ReplicaSet kubernetes-dashboard/kubernetes-dashboard-7b76878768
E0715 16:11:46.862517    2619 replica_set.go:535] sync "kubernetes-dashboard/kubernetes-dashboard-7b76878768" failed with pods "kubernetes-dashboard-7b76878768-" is forbidden: no providers available to validate pod request
I0715 16:11:46.862552    2619 event.go:278] Event(v1.ObjectReference{Kind:"ReplicaSet", Namespace:"kubernetes-dashboard", Name:"kubernetes-dashboard-7b76878768", UID:"663e530b-8726-4619-88c8-d8374a6226f2", APIVersion:"apps/v1", ResourceVersion:"1878", FieldPath:""}): type: 'Warning' reason: 'FailedCreate' Error creating: pods "kubernetes-dashboard-7b76878768-" is forbidden: no providers available to validate pod request
```

Additionally, I am unable to create pods myself:
```
$ cat ubuntu.yaml
apiVersion: v1
kind: Pod
metadata:
  name: ubuntu
  namespace: default
spec:
  containers:
  - image: ubuntu
    command:
      - sleep
      - "3600"
    imagePullPolicy: IfNotPresent
    name: ubuntu
  restartPolicy: Always

$ kubectl apply -f ubuntu.yaml
Error from server (Forbidden): error when creating "ubuntu.yaml": pods "ubuntu" is forbidden: no providers available to validate pod request
```

joedborg: It is working when I set it post default 1.18, but it doesn't if I deploy from scratch, due to the scheduler not being able to talk to the API:
`2020-07-15T16:20:07Z kube-scheduler.daemon[20242]: E0715 16:20:07.401045   20242 reflector.go:178] k8s.io/client-go/informers/factory.go:135: Failed to list *v1.Service: Get http://127.0.0.1:8080/api/v1/services?limit=500&resourceVersion=0: dial tcp 127.0.0.1:8080: connect: connection refused`

reactive/kubernetes_master.py
```diff
@@ -1944,11 +1964,11 @@ def configure_apiserver():
 
     admission_plugins = [
         'PersistentVolumeLabel',
+        'PodSecurityPolicy',
+        'NodeRestriction'
     ]
 
     auth_mode = hookenv.config('authorization-mode')
-    if 'Node' in auth_mode:
```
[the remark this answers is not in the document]
Cynerva: I think this is safe. The NodeRestriction admission controller is documented [here](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction). It's just going to prevent kubelet from modifying Node or Pod resources that belong to other kubelets. Kubelet should never try to do that anyway.

reactive/kubernetes_master.py
```diff
@@ -1813,21 +1850,21 @@ def build_kubeconfig():
 
         # make a copy in a location shared by kubernetes-worker
         # and kubernete-master
-        create_kubeconfig(kubeclientconfig_path, local_server, ca_crt_path,
+        create_kubeconfig(kubeclientconfig_path, public_server, ca_crt_path,
```
Cynerva, line 1853: I'm not convinced switching to `public_server` across the board is necessary or a good idea. This is going to force all local API traffic to go through external loadbalancers. What are you seeing that makes this necessary?

johnsca: Agreed.  `local_server` is using the ingress address for the `kube-api-endpoint` relation, which seems like it should be fine.

joedborg: Yep, I've looked over this again.  I was conflating loopback address vs local address, this works fine with local_server.

reactive/kubernetes_master.py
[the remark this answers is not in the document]
joedborg: Thanks!

joedborg:
Requires https://github.com/juju-solutions/interface-kube-control/pull/31to be merged first

reactive/kubernetes_master.py
```diff
@@ -1154,7 +1183,7 @@ def create_tokens_and_sign_auth_requests():
             kubelet_token = get_token(username)
         kube_control.sign_auth_request(request[0], username,
                                        kubelet_token, proxy_token,
-                                       client_token)
+                                       scheduler_token, client_token)
```
Cynerva, line 1186: I see no need to send the system:kube-scheduler token to kubernetes-worker.

reactive/kubernetes_master.py
```diff
@@ -2348,6 +2421,9 @@ def poke_network_unavailable():
     discussion about refactoring the affected code but nothing has happened
     in a while.
     """
+    public_address, public_port = kubernetes_master.get_api_endpoint()
+    public_server = 'https://{0}:{1}'.format(public_address, public_port)
```
Cynerva, line 2425: I would prefer the "local address/server" here - the ingress address of kube-api-endpoint. For the same reason that we prefer it for kube-scheduler, kube-controller-manager, etc. No need to send API traffic over the network to loadbalancers if we don't have to.

reactive/kubernetes_master.py
```diff
@@ -2363,7 +2439,7 @@ def poke_network_unavailable():
 
     for node in nodes:
         node_name = node['metadata']['name']
-        url = 'http://localhost:8080/api/v1/nodes/{}/status'.format(node_name)
+        url = '{}/api/v1/nodes/{}/status'.format(public_server, node_name)
         with urlopen(url) as response:
```
Cynerva, line 2443: This function will need some more work. The requests are going to fail - they do not include a token. I would use the admin token.

reactive/kubernetes_master.py
joedborg: +1

Cynerva:
LGTM

joedborg:
Tested deployed as an overlay to CK 1.18 and as an upgrade path (with and without setting `Node,RBAC`).  All look good.

reactive/kubernetes_master.py
```diff
@@ -2250,11 +2250,19 @@ def configure_scheduler():
     scheduler_opts['profiling'] = 'false'
     scheduler_opts['config'] = kube_scheduler_config_path
 
+    scheduler_ver = get_version('kube-scheduler')
```
kwmonroe, line 2253: FYI, we needed to switch the api version based on the k8s version.

kubescheduler.config.k8s.io/v1alpha1 was removed in 1.19 by kubernetes/kubernetes#89298.
kubescheduler.config.k8s.io/v1alpha2 was added in 1.18 by kubernetes/kubernetes#87628.

PR #105, opened 2020-07-24
Apply CSI encryption config

joedborg:
Needed for Ceph CSI 2.1.2.   Tested on CK1.18 and addons from https://github.com/charmed-kubernetes/cdk-addons/pull/190

Fixes: https://bugs.launchpad.net/cdk-addons/+bug/1867940

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1482,6 +1482,22 @@ def ceph_storage_pool():
     set_state('kubernetes-master.ceph.pool.created')
 
 
+@when('leadership.is_leader')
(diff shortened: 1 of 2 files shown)


johnsca:
I agree with @kwmonroe in thinking it's a bit strange to have this one template happen outside of cdk-addons. It looks like we might want to populate it with info from Vault at some point in the future, but presumably it would be sufficient to add a `ceph-kms-config` snap config to pass in JSON to put in the template, defaulting it to `{}`.

joedborg:
Rolled into charmed-kubernetes/cdk-addons#190

PR #106, opened 2020-07-28
support webhook authentication

kwmonroe:
Fixes:
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1878455
https://bugs.launchpad.net/charm-kubernetes-worker/+bug/1887810

We need to remove --basic and --token auth from the k8s apiserver in order to pass the CIS K8s benchmark.  This PR gets us there by:
- creating a webhook auth service that can authenticate against k8s secrets
- converting known_tokens to k8s secrets
- adjusting the k8s-master charm to use the new webhook


templates/cdk.master.auth-webhook.py
```diff
@@ -0,0 +1,135 @@
...
+    app.logger.info('Checking AWS')
+    return False
+
+
+def check_keystone(token_review):
+    app.logger.info('Checking Keystone')
+    return False
+
+
+@app.route('/', methods=['POST'])
```
ktsakalozos, line 93: You may want to add some versioning in the url so we can assist upgrades, beta testing etc

templates/cdk.master.auth-webhook.py
```diff
@@ -0,0 +1,135 @@
...
+        valid = False
+
+    if valid:
+        # Make the request unauthenticated by deafult
+        req['status'] = {'authenticated': False}
+    else:
+        app.logger.info('Invalid request: {}'.format(req))
+        return ''  # flask needs to return something that isn't None
+
+    app.logger.debug('REQ: {}'.format(req))
```
johnsca, line 123: Should mask the password / token field from the request in the log.

kwmonroe:
/hold

The auth-webhook kubeconfig isn't right.  /me needs to see where in the 42 forced pushes the right way got clobbered.

kwmonroe:
/unhold

Between moving service account actions out (#108) and fixing master merge conflicts, I lost changes to the auth-webhook and k8s-master handlers that made fresh deployments possible. These have now been committed, so we're g2g again.

templates/cdk.master.auth-webhook.py
```diff
@@ -0,0 +1,255 @@
...
+    try:
+        valid = True if (req['kind'] == 'TokenReview' and
+                         req['spec']['token']) else False
+    except (KeyError, TypeError):
+        valid = False
+
+    if valid:
+        app.logger.debug('REQ: {}'.format(req))
+    else:
+        app.logger.info('Invalid request: {}'.format(req))
```
johnsca, line 227: Still need to mask out the password / tokens from the logs.

templates/cdk.master.auth-webhook.py
```diff
@@ -0,0 +1,255 @@
...
+        or check_keystone(req)
+        {%- endif %}
+        {%- if custom_authn_endpoint %}
+        or check_custom(req)
+        {%- endif %}
+       ):
+        # Successful checks will set auth and user data in the 'req' dict
+        app.logger.debug('ACK: {}'.format(req))
+    else:
+        app.logger.debug('NAK: {}'.format(req))
```
johnsca, line 249: Also need to mask out the password / tokens from the log.  Something like:

```suggestion
from copy import deepcopy
...
    log_req = deepcopy(req)
    log_req['spec']['token'] = '******'
    ...
    if (...):
        app.logger.debug('ACK: {}'.format(log_req))
    else:
        app.logger.debug('NAK: {}'.format(log_req))
```

actions/user-delete
```diff
@@ -0,0 +1 @@
+user-create
```
johnsca, line 1: Bike shedding: Since you're symlinking anyway, you could make `user-create` a symlink as well and change the implementation filename to `manage_users.py` so that the syntax highlighting would work on the webview (plus it's a tiny bit more accurate, since the file contains more than just the `user_create` function).

reactive/kubernetes_master.py
```diff
@@ -281,8 +283,14 @@ def post_series_upgrade():
...
 
 def add_rbac_roles():
-    '''Update the known_tokens file with proper groups.'''
+    '''Update the known_tokens file with proper groups.
 
+    DEPRECATED: Once known_tokens are migrated, group data will be stored in K8s
+    secrets. Do not use this function after migrating to authn with secrets.
+    '''
+    if is_flag_set('kubernetes-master.token-auth.migrated'):
+        hookenv.log('Known tokens have migrated to secrets. Skipping group changes')
```
johnsca, line 292: Wouldn't hitting this case imply that it's a charm code error, likely caused by a new change introducing a regression?  It feels like this should perhaps be more fatal.

reactive/kubernetes_master.py
```diff
@@ -570,11 +578,11 @@ def setup_leader_authentication():
     # Try first to fetch data from an old leadership broadcast.
     if not get_keys_from_leader(keys) \
             or is_state('reconfigure.authentication.setup'):
-        if not os.path.isfile(basic_auth):
-            kubernetes_master.deprecate_auth_file(basic_auth)
+        kubernetes_master.deprecate_auth_file(basic_auth)
+        set_flag('kubernetes-master.basic-auth.migrated')
```
johnsca, line 582: It confused me at first that the upgrade handler is checking the return value of the deprecate function but this is not.  But I don't think it actually matters in practice.

reactive/kubernetes_master.py
kwmonroe: Not sure I follow.  This function ensures groups are correctly added to `known_tokens.csv`, which they still are during `upgrade-charm` for clusters < 1.19 -- in fact, `add_rbac_roles` is *only* called on upgrade. Once `k_t.csv` is migrated to k8s secrets, there will be no reason to add groups, as that information will have already been stored in the secret.

Once secrets have been around for 3 releases, `add_rbac_roles()` should be removed.

reactive/kubernetes_master.py
kwmonroe: Technically, the upgrade handler is checking the return of *migrate_auth_file*, not *deprecate_auth_file*.  The thinking here is that you'll only get to this code if:

1. Brand new deployment with no leader keys; in this case, we always want to write the deprecated file and, more importantly, set the `.migrated` flag so we know to check for secrets from the get-go.
2. Upgraded deployment at which point `upgrade-charm` will have already done the migration, so checking `deprecate` status seems moo, like a cow's opinion.

kwmonroe:
ugh, my legacy will forever have `objuscate` in my commit history.

reactive/kubernetes_master.py
johnsca: Ah, gotcha.  I think I was confused by the last sentence of the deprecation notice.

reactive/kubernetes_master.py
johnsca: Right you are.  My mistake.

lib/charms/layer/kubernetes_master.py
```diff
@@ -141,13 +149,170 @@ def get_cephfs_fsname():
...
-    In 1.19+, file-based authentication was deprecated in favor of cert
+    In 1.19+, file-based authentication was deprecated in favor of webhook
     auth. Write out generic files that inform the user of this.
     """
     csv_file = Path(auth_file)
     csv_file.parent.mkdir(exist_ok=True)
+
+    csv_backup = Path('{}.{}'.format(csv_file, AUTH_BACKUP_EXT))
+    if csv_file.exists() and not csv_backup.exists():
+        csv_file.rename(csv_backup)
```
johnsca, line 160: A brand new bootstrap ends up with a `known_tokens.csv.pre-migration` file with an incorrect admin user token in it.  Not a big deal, but could be a bit confusing, so it might be nice to avoid it if we can.

lib/charms/layer/kubernetes_master.py
```diff
@@ -141,13 +149,170 @@ def get_cephfs_fsname():
...
+
+
+def create_secret(token, username, user, groups=None, ns='auth-webhook'):
+    sani_name = re.sub('[^0-9a-zA-Z]+', '-', user)
+    secret_id = '{}-secret'.format(sani_name)
+    delete_secret(secret_id, ns=ns)
+    # The authenticator expects tokens to be in the form user::token
+    token_delim = '::'
+    if token_delim not in token:
+        token = '{}::{}'.format(user, token)
```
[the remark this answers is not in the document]
johnsca: Ah, I guess the switch from basic auth to tokens is going to require refreshing kubectl config files anyway, so maybe this isn't actually a concern.  Feel free to disregard this.

templates/cdk.master.auth-webhook.service
```diff
@@ -0,0 +1,21 @@
...
+
+[Service]
+User=root
+WorkingDirectory=/root/cdk/auth-webhook
+ExecStart={{ charm_dir }}/../.venv/bin/gunicorn \
+    --bind {{ host }}:{{ port }} \
+    --capture-output \
+    --certfile /root/cdk/server.crt \
+    --disable-redirect-access-to-syslog \
+    --error-logfile auth-webhook.log \
```
johnsca, line 13: There seems to be a fair amount of periodic activity that ends up in the log file.  We probably need to set up log rotation and we might want to consider separating the error and access logs, or changing the default log level (should it be configurable somehow?).

tests/test_kubernetes_master_actions.py
```diff
@@ -0,0 +1,77 @@
+import base64
+import json
+import pytest
+import sys
+from unittest import mock
+
+
+charms = mock.MagicMock()
+sys.modules['charms'] = charms
+sys.modules['charms.layer'] = charms.layer
```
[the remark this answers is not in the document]
johnsca: Ah, it seems this was working around an issue where `import_layer_libs` was not patched.  That's fixed in https://github.com/juju-solutions/charms.unit_test/pull/12 and I tested it against this branch and seems to allow these lines to be removed.

lib/charms/layer/kubernetes_master.py
```diff
@@ -141,13 +149,170 @@ def get_cephfs_fsname():
...
+
+    if kubernetes_common.kubectl_success(
+      '-n', ns, 'create', 'secret', 'generic', secret_id,
+      "--from-literal=username={}".format(username),
+      "--from-literal=groups={}".format(groups or ''),
+      "--from-literal=password={}".format(token)):
+        hookenv.log("Created secret for {}".format(username))
```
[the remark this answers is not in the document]
kwmonroe: What do you think about adjusting k8s-common [kubectl](https://github.com/charmed-kubernetes/layer-kubernetes-common/blob/master/lib/charms/layer/kubernetes_common.py#L236) to include a `kwargs` check for something like `mask_arg`, where a matching arg would just be replaced with `********` in the hookenv log?

I see the point you're making with running a subproc `kubectl` command, but maybe there are other k8s-common users that could benefit from masking values from the log?

tests/test_kubernetes_master_actions.py
```diff
@@ -0,0 +1,77 @@
...
+    secret_id = '{}-secret'.format(user)
+    test_data = {
+        user: secret_id
+    }
+
+    mock_get.return_value = user
+    with mock.patch('actions.user_actions.user_list',
+                    return_value=test_data):
+        user_actions.user_create()
+        assert user_actions.action_fail.called
```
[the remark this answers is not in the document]
kwmonroe: Correct, that's because `lib.k8s-master.create_secret` is already unit tested in the k8s-master lib unit tests.  That said, there is a difference between the charm and the action here; this action calls `k8s_master.create_secret` *as well as* `k8s_common.create_kubeconfig`, but i don't think this action should unit test those funcs.

Feels better to me for those tests to happen in their respective modules.  LMK if i'm missing a scenario other than "creating an existing user == failure" that should be tested here.

tests/test_kubernetes_master_actions.py
```diff
@@ -0,0 +1,77 @@
...
+    secret_id = '{}-secret'.format(user)
+    test_data = {
+        user: secret_id
+    }
+
+    mock_get.return_value = 'testuser2'
+    with mock.patch('actions.user_actions.user_list',
+                    return_value=test_data):
+        user_actions.user_delete()
+        assert user_actions.action_fail.called
```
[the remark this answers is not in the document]
kwmonroe: Same.

lib/charms/layer/kubernetes_master.py
```diff
@@ -141,13 +149,170 @@ def get_cephfs_fsname():
...
+        del record['groups']
+
+    with known_tokens.open('w') as f:
+        csv.DictWriter(f, csv_fields, lineterminator='\n').writerows(
+            tokens_by_user.values())
+
+
+def create_secret(token, username, user, groups=None, ns='auth-webhook'):
+    sani_name = re.sub('[^0-9a-zA-Z]+', '-', user)
+    secret_id = '{}-secret'.format(sani_name)
```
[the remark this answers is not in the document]
kwmonroe: Good point.  I wanted to make it easy to spot the CK secrets when listing all of them, but we'll get that when setting a custom type.

reactive/kubernetes_master.py
```diff
@@ -209,27 +211,27 @@ def check_for_upgrade_needed():
...
-                    continue
-                else:
-                    setup_tokens(*row)
-            except IndexError:
-                pass
-        kubernetes_master.deprecate_auth_file(basic_auth)
-
-        set_flag('kubernetes-master.basic-auth.migrated')
+    # File-based auth is gone in 1.19; ensure any entries in basic_auth.csv are
+    # added to known_tokens.csv, and any known_tokens entries are created as secrets.
```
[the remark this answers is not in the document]
kwmonroe: oooof!  excellent catch.  luckily your future commentary will fix this -- that is, instead of namespacing secrets, we'll set them up with a custom type.

actions/user_actions.py
```diff
@@ -0,0 +1,112 @@
...
+        '-n', NS, 'get', 'secrets', '-o', 'json').decode('UTF-8')
+    secrets = json.loads(output)
+    users = {}
+    if 'items' in secrets:
+        for secret in secrets['items']:
+            try:
+                secret_id = secret['metadata']['name']
+                username_b64 = secret['data']['username'].encode('UTF-8')
+            except (KeyError, TypeError):
+                # CK secrets will have populated 'data', but not all secrets do
```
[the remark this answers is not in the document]
kwmonroe: This is great -- i didn't actually think to check if other types would be more fitting.  This is doubly great because i namespaced all our secrets just to keep them together.  I can drop all the NS code and just `juju.is/secret` or whatever.

lib/charms/layer/kubernetes_master.py
kwmonroe: Yeah, this happens because sometimes [password_changed](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/reactive/kubernetes_master.py#L525) is called before `setup_leader_auth`.  I haven't been able to come up with a way around this without possibly losing the `config.changed.password` when config is changed post-deployment.

lib/charms/layer/kubernetes_master.py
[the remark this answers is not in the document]
johnsca: That sounds like a great idea.

templates/cdk.master.auth-webhook.service
kwmonroe: The debug level is set in the systemd service file.  We could add a config opt to k8s-master to control this, or document how users can modify that service file on-disk for all masters.  The former is nicer, but adds to our growing list of config opts.  Whatcha think?

tests/test_kubernetes_master_actions.py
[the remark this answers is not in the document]
johnsca: My issue with that is that almost the entire function is below that `user in users` check so doesn't get covered. 
 `k8s_common.create_kubeconfig` is already mocked out, and it should be easy to temporarily mock out `layer.k8s_master` so that you can cover the rest of that function without it doing any external function calls.

tests/test_kubernetes_master_actions.py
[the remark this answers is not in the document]
johnsca: Same.

actions/user_actions.py
```diff
@@ -0,0 +1,112 @@
...
+    action_set,
+    action_fail,
+    action_name
+)
+from charms import layer
+
+os.environ['PATH'] += os.pathsep + os.path.join(os.sep, 'snap', 'bin')
+
+# Default namespace for all CK secrets
+NS = 'auth-webhook'
```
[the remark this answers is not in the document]
kwmonroe: I dropped namespaced secrets in favor of your most excellent custom type, so this is now moo.

templates/cdk.master.auth-webhook.service
kwmonroe: We've decided to keep debug hard coded (at least for now) and implement log rotation to address space issues.

lib/charms/layer/kubernetes_master.py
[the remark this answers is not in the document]
kwmonroe: Maybe a task for the future, but now that we're creating secrets with a manifest, we don't leak secrets in the logs..

lib/charms/layer/kubernetes_master.py
kwmonroe: I was able to work around this by reconfiguring auth and *not* writing a token in `password_changed`.  We still get the password changed, we just write it as a secret instead of to a file during new deployments.

lib/charms/layer/kubernetes_master.py
```diff
@@ -141,13 +149,170 @@ def get_cephfs_fsname():
...
+    delete_secret(secret_id, ns=ns)
```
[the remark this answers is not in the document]
kwmonroe: This was happening because we always called `delete_secret` early in `create_secret` -- we did that because we were using `kubectl create` which would have failed if the user already existed... Now that we're using `kubectl apply`, we can just stomp over existing secrets willy nilly.

Before:
```bash
  results:
    Stderr: |
      Error from server (NotFound): secrets "t1" not found
    Stdout: |
      Cluster "juju-cluster" set.
      Property "users" unset.
      User "t1" set.
      Context "juju-context" created.
      Switched to context "juju-context".
    kubeconfig: juju scp kubernetes-master/0:/home/ubuntu/t1-kubeconfig .
    msg: User "t1" created.
    users: admin, system:kube-controller-manager, system:kube-proxy, system:node:ip-172-31-0-211,
      system:node:ip-172-31-11-121, system:node:ip-172-31-22-158, system:kube-scheduler,
      system:monitoring, t1
  status: completed
```

Now:
```bash
  results:
    Stdout: |
      Cluster "juju-cluster" set.
      Property "users" unset.
      User "t2" set.
      Context "juju-context" created.
      Switched to context "juju-context".
    kubeconfig: juju scp kubernetes-master/0:/home/ubuntu/t2-kubeconfig .
    msg: User "t2" created.
    users: admin, system:kube-controller-manager, system:kube-proxy, system:node:ip-172-31-0-211,
      system:node:ip-172-31-11-121, system:node:ip-172-31-22-158, system:kube-scheduler,
      system:monitoring, t1, t2
  status: completed

```

Given this change, I think it would be useful to leave stderr in-tact if delete fails.

PR #111, opened 2020-08-24
ensure new service secrets include expected groups

kwmonroe:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1892764

As noted in the [upstream docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#core-component-roles), rbac enabled clusters have a few default cluster roles that grant access for k8s components.  k8s will automatically bind the necessary cluster role based on the username, but since we support `groups` in our secrets, we should make sure this field is an accurate depiction of our rbac-enabled cluster when creating new secrets.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1185,21 +1185,23 @@ def create_tokens_and_sign_auth_requests():
     # create proper secrets.
     controller_manager_token = get_token('system:kube-controller-manager')
     if not controller_manager_token:
-        setup_tokens(None, 'system:kube-controller-manager',
-                     'kube-controller-manager')
+        setup_tokens(None, 'system:kube-controller-manager', 'kube-controller-manager',
+                     'system:kube-controller-manager')
 
     proxy_token = get_token('system:kube-proxy')
     if not proxy_token:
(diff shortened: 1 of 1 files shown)


Cynerva:
I don't understand what this fixes.

> ensure new service secrets include expected groups

Who or what is expecting these groups? I see no mention in the upstream docs of a `system:node-proxier` group. The docs mention a ClusterRole named `system:node-proxier`, but a ClusterRole is not a group, and the `system:node-proxier` ClusterRoleBinding does not appear to reference a `system:node-proxier` group. Where are these groups coming from?

kwmonroe:
> I don't understand what this fixes.

Oof, that's because it doesn't actually fix anything.  We set the `system:masters` group (a valid *group*) when setting up the admin token, and pass that back in the auth-webhook TokenReview response.  My only excuse for this PR is that i saw `system:foo` in the rbac components docs and thought, "oh, we should do the same thing that we do for `system:masters`".

> a ClusterRole is not a group

Yup. Worse yet, I noted in the desc that k8s would bind cluster roles to users, but it still didn't click that roles != groups.  My bad, and thanks for the sanity.  I'm closing this.

PR #112, opened 2020-08-24
ensure valid default when ceph fsname is unavailable

kwmonroe:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1892767

We append ceph[fsname] to the arg string when configuring cdk-addons. Make sure this is an empty string when cephFs is not available.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1368,10 +1368,9 @@ def configure_cdk_addons():
         default_storage = hookenv.config('default-storage')
         if kubernetes_master.query_cephfs_enabled():
             cephFsEnabled = "true"
-            ceph['fsname'] = kubernetes_master.get_cephfs_fsname()
+            ceph['fsname'] = kubernetes_master.get_cephfs_fsname() or ''
         else:
             cephFsEnabled = "false"
-            ceph['fsname'] = ''
     else:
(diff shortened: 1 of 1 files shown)


Cynerva:
I don't think this fixes the issue.

> Make sure this is an empty string when cephFs is not available.

That's already handled here, right?

https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/072f8e5e9cd7a56db742dc39310033737653ab73/reactive/kubernetes_master.py#L1418

Looking at the failure, it looks more to me like fsname *was* set in the ceph dict, but was set to None. The only place that could have happened is here:

https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/072f8e5e9cd7a56db742dc39310033737653ab73/reactive/kubernetes_master.py#L1371

kwmonroe:
> That's already handled here, right?
> 
> https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/072f8e5e9cd7a56db742dc39310033737653ab73/reactive/kubernetes_master.py#L1418

Yes, yes it is.  I've updated this PR so we'll set an empty string if fsname comes back as None (versus the case where `.get` would set a default value).

Cynerva:
LGTM

PR #119, opened 2020-09-02
Make auth-webhook listen on 127.0.0.1

johnsca:
Occasionally on certain clouds, `socket.gethostname()` will return a value which cannot be resolved for some reason. Since the auth-webhook tries to bind to that, it fails to start and blocks the cluster. And since it only needs to be contacted by the API server running on the same machine, this instead binds it to localhost.

Fixes [lp:1893995](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1893995)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -983,7 +983,7 @@ def register_auth_webhook():
     #   https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18
     context = {'api_ver': 'v1beta1',
                'charm_dir': hookenv.charm_dir(),
-               'host': '127.0.0.1',
+               'host': get_ingress_address('kube-api-endpoint'),
                'pidfile': 'auth-webhook.pid',
                'port': 5000,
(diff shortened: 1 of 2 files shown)


johnsca:
Tested on AWS with a cluster where the instances couldn't resolve their hostname.  Tested from outside the cluster, talking to the kubeapi-load-balancer with both the default admin user and a custom user.

Cynerva:
I would expect 127.0.0.1 to lead to TLS errors given that the webhook appears to serve with /root/cdk/server.crt, which does not include 127.0.0.1 in its SANs. For local HTTPS connections, we usually use `get_ingress_address('kube-api-endpoint')` instead.

johnsca:
The config currently specifies `insecure-skip-tls-verify: true` so TLS isn't validated anyway, but I agree that validating TLS is probably better than binding to localhost (though MITM'ing localhost would presumably require root anyway).

kwmonroe (approved):

PR #126, opened 2020-09-24
Add support for DNS provider as a charm

johnsca:
Add support for the DNS provider (i.e., CoreDNS) to be related as a charm (e.g., one running in the cluster).

Part of [lp:1858007](https://bugs.launchpad.net/charmed-kubernetes-bundles/+bug/1858007)

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1216,7 +1216,7 @@ def send_cluster_dns_detail(kube_control):
             except CalledProcessError:
                 hookenv.log("DNS addon service not ready yet")
                 return
-        kube_control.set_dns(53, dns_domain, dns_ip, dns_enabled)
+            kube_control.set_dns(53, dns_domain, dns_ip, dns_enabled)
 
 
 def create_tokens_and_sign_auth_requests():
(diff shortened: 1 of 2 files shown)


reactive/kubernetes_master.py
```diff
@@ -1194,21 +1194,29 @@ def etcd_data_change(etcd):
...
-    try:
-        dns_provider = get_dns_provider()
-    except InvalidDnsProvider:
-        hookenv.log(traceback.format_exc())
-        return
-    dns_enabled = dns_provider != 'none'
-    dns_domain = hookenv.config('dns_domain')
-    dns_ip = None
-    if dns_enabled:
+    dns_provider = endpoint_from_flag('dns-provider.available')
```
kwmonroe, line 1197: This is knowable without `cdk-addons.configured`, right?  I see that this will allow a related dns charm to take precedent over whatever we deliver in cdk-addons, but i'm concerned that a dns charm *is coming*, but isn't available yet.  In that case, masters/workers will do the `get_dns_provider` rigmarole only to have to do it again once that thing is available.

Would it be possible / useful to use `goal-state` here to see if a dns relation is coming, and if it is, wait for it; if it's not, continue as if cdk-addons was providing dns like it has in the past.

johnsca: On further reflection, this is moot.  The relation won't exist until after the cluster is fully up, because it's only after that point that you can add the cluster to your controller as a K8s cloud and then deploy the CoreDNS K8s charm into it.  The initial deployment of the cluster will have to be done with the `dns-provider` config set to `none` (or have the value changed to `none` after the deployment is up, if desired).

kwmonroe (approved):

PR #127, opened 2020-09-25
Don't require full Keystone creds from relation

johnsca:
The charm actually only cares about the auth API URL but because there's not a relation for just that from Keystone, it instead uses the relation for requesting a new user (keystone-credentials). The interface layer does provide a flag for the case where just the API endpoint info is available, so it's better to use that instead of blocking on the full credentials being available.

This fixes part of [lp:1834164][] where the integrator charm might not have credentials in the format required by the relation but can still at least provide the API endpoint info.

[lp:1834164]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1834164

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -139,9 +139,9 @@ register_trigger(when='kubernetes-master.ceph.configured',
                  set_flag='cdk-addons.reconfigure')
 register_trigger(when_not='kubernetes-master.ceph.configured',
                  set_flag='cdk-addons.reconfigure')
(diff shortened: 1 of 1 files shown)


johnsca:
For reference, the `keystone-credentials.available` flag is set [here](https://github.com/openstack/charm-interface-keystone-credentials/blob/master/requires.py#L52) based on [this set of data](https://github.com/openstack/charm-interface-keystone-credentials/blob/master/requires.py#L72-L85) being available.

kwmonroe (approved):
LGTM

PR #128, opened 2020-10-07
Generate project scoped Keystone token for Dashboard.

freyes:
When a user doesn't have a default project set the current kube-keystone.sh
script generates invalid tokens. This patch explicitly requests a scoped token
honouring the OS_* environment variables set.

Fixes [lp:1893214](https://pad.lv/1893214)


Cynerva:
LGTM. I tested and confirmed that with this change, I can now run get_keystone_token, put the token into kubernetes-dashboard, and successfully authenticate as a user that has no default project set. Thank you!

PR #129, opened 2020-11-09
Temporary snap store failures shouldn't trigger new cohorts

johnsca:
Small addition to [lp:1893220][] to ensure that occasional temporary request failures to the snap store (I occasionally see 503 errors) don't lead to the charms trying to create a new set of cohorts. The request error is already handled in the `get_snap_rev` function, but results in an empty dict being returned rather than the current list of revisions, which would then trigger a new cohort rollout (which might then fail due to the transient snap store errors).

[lp:1893220]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1893220


Cynerva:
LGTM :D

PR #131, opened 2020-11-18
skip invalid known_tokens.csv entries

kwmonroe:
Skip known_tokens.csv entries that don't look like a valid token.

Fixes: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1904767

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -298,31 +298,39 @@ def add_rbac_roles():
     with open(tokens_fname, 'w') as ftokens:
         with open(tokens_backup_fname, 'r') as stream:
             for line in stream:
+                if line.startswith('#'):
+                    continue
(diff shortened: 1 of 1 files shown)


johnsca (approved):
LGTM, though I'm confused how we hit this. I take it it wasn't strictly related to the upgrade process and a commented / invalid line got in there in some other way?

kwmonroe:
> I'm confused how we hit this.

This *could* happen if you didn't have a `basic_auth.csv` and you manually stuffed junk into `known_tokens.csv`.  This *did* happen because CI tried to test an upgrade from `stable` charms to `candidate` charms.  Unfortunately, the `candidate` charms were old, so this ended up being a downgrade test.  Downgrading to a charm rev that came before the migration bits would error as described in the bug.

PR #132, opened 2020-11-24
Fix followers advertising stale creds to kube-control

Cynerva:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1905058
Depends on https://github.com/juju-solutions/interface-kube-control/pull/32

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -283,6 +283,11 @@ def post_series_upgrade():
     remove_state('kubernetes-master.components.started')
 
 
+@hook('leader-elected')
+def leader_elected():
+    clear_flag('authentication.setup')
+
+
 def add_rbac_roles():
     '''Update the known_tokens file with proper groups.
 
@@ -648,6 +653,12 @@ def setup_non_leader_authentication():
 
     if any_file_changed(keys):
         remove_state('kubernetes-master.components.started')
+
+    # Clear stale creds from the kube-control relation so that the leader can
+    # assume full control of them.
+    kube_control = endpoint_from_flag('kube-control.connected')
+    if kube_control:
+        kube_control.clear_creds()
 
     remove_state('kube-control.requests.changed')
     set_state('authentication.setup')


Cynerva:
I tested this with the following steps:

1. Deployed CK 1.18+ck1, waited for it to settle.
1. Observed kubernetes-master/1 is the initial leader.
1. Forced a new leader election:
   1. `juju ssh kubernetes-master/1 sudo service jujud-unit-kubernetes-master-1 stop`.
   1. Waited for Juju to elect kubernetes-master/0 as the new leader.
   1. `juju ssh kubernetes-master/1 sudo service jujud-unit-kubernetes-master-1 start`.
1. Forced the new leader to advertise creds.
   1. `juju run -u kubernetes-master/0 charms.reactive clear_flag authentication.setup '&&' hooks/update-status`.
1. Observed both kubernetes-master/0 and kubernetes-master/1 are advertising creds.
   1. `juju run -u kubernetes-worker/0 -- 'relation-get -r $(relation-ids kube-control) - kubernetes-master/0'`
   1. `juju run -u kubernetes-worker/0 -- 'relation-get -r $(relation-ids kube-control) - kubernetes-master/1'`
1. Upgraded charm to a locally built charm with the changes from this PR.
1. Verified that kubernetes-master/0 was advertising new creds.
1. Verified that kubernetes-master/1 *stopped* advertising creds.
1. Verified that kubernetes-workers settled properly and were using the new creds as they should.

Without these changes, the workers never would have gotten the new creds because kubernetes-master/1 would still have been advertising old creds on the relation. The kubernetes-master unit with the highest unit ID happens to win because of the [merge logic in kube-control](https://github.com/juju-solutions/interface-kube-control/blob/ea90ca566da63750ab90bc1443c77d912abf6d58/requires.py#L56-L58).

After the above test sequence, I also tested leader re-elections a couple more times to ensure good behavior. I was able to verify multiple times that the new leader immediately started advertising credentials, and the old leader immediately stopped advertising them.

DomFleischmann (approved):
LGTM, besides the comment that @kwmonroe left in https://github.com/juju-solutions/interface-kube-control/pull/32 . I'm also curious about why we leave an empty value instead of removing the key.

Cynerva:
Thanks. I've since commented in https://github.com/juju-solutions/interface-kube-control/pull/32. The short answer is the code there *does* remove the key from the relation completely, and I could not find a cleaner way to do it.

PR #136, opened 2021-01-11
Set blocked state if CA is missing

mkalcok:
I took a same approach as  [etcd charm](https://github.com/charmed-kubernetes/layer-etcd/blob/39154a235f7e6996b94d069938752c98055d0f65/reactive/etcd.py#L128) and added check for missing `certificates.available` flag.

I have the same thing prepared for the kubernetes-worker, but I'm gonna open this PR first to hammer out any possible issues.

LP: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1868541

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -703,11 +703,6 @@ def set_app_version():
     hookenv.application_version_set(version.split(b' v')[-1].rstrip())
 
 
-@when_not('certificates.available')
-def missing_certificates_notice():
-    hookenv.status_set('blocked', 'Missing relation to certificate authority.')
-
-
 @hookenv.atstart
 def check_vault_pending():
     try:
@@ -736,6 +731,11 @@ def set_final_status():
         hookenv.status_set('blocked',
                            'Series upgrade in progress')
         return
+
+    if not is_flag_set('certificates.available'):
+        hookenv.status_set('blocked', 'Missing relation to certificate authority.')
+        return
+
     if is_flag_set('kubernetes-master.secure-storage.failed'):
         hookenv.status_set('blocked',
                            'Failed to configure encryption; '
@@ -896,10 +896,6 @@ def set_final_status():
     if is_leader and ks and \
        not is_flag_set('kubernetes-master.keystone-policy-handled'):
         hookenv.status_set('waiting', 'Waiting to apply keystone policy file.')
-        return
-
-    if not is_flag_set('certificates.available'):
-        missing_certificates_notice()
         return
 
     hookenv.status_set('active', 'Kubernetes master running.')
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -703,6 +703,11 @@ def set_app_version():
     hookenv.application_version_set(version.split(b' v')[-1].rstrip())
 
 
+@when_not('certificates.available')
+def missing_certificates_notice():
+    hookenv.status_set('blocked', 'Missing relation to certificate authority.')
+
+
 @hookenv.atstart
 def check_vault_pending():
     try:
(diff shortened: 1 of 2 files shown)


reactive/kubernetes_master.py
```diff
@@ -893,6 +898,10 @@ def set_final_status():
         hookenv.status_set('waiting', 'Waiting to apply keystone policy file.')
         return
 
+    if not is_flag_set('certificates.available'):
+        missing_certificates_notice()
```
r-f-g, line 902: With this approach, `get_flags()` function is called twice (in decorator `when_not` and in function `is_flag_set`). Do you want to use the `missing_certificates_notice` function elsewhere?

mkalcok: Yeah, I'm not particularly happy about this either, but as you can see, that's the general purpose of this `@atexit` function. You need to stop its execution otherwise it sets status to `active` at the end. Without this `if`, the flow would go like this:
* `certificates.available flag changes
* `missing_certificates_notice()` is executed
* `blocked` state is set
* `@atexit` function is executed
* `active` status is set

r-f-g: Yes, I know that this `if` is needed, but the `missing_certificates_notice` function isn't?

- `certificates.available` flag changes
- `@atexit` function is executed
- `blocked` status is set (with `if` and `hookenv.status_set(...)`)

The correct question is whether the `atexit` function will be triggered when the `certificates.available` (or another) flag changes. Do you know if yes?

mkalcok: Oh, you meant it like that...I don't think `@atexit` would be executed without explicit hook reacting to the flag change. According to the documentation `atexit: Schedule a callback to run on successful hook completion.` I can give it a shot though.

mkalcok: As suspected, without the explicit hook the status wont change to `blocked`, at least not immediately. Eventually some other hook runs and the status is changed but that could take up to 5 minutes (`update-status`) and I don't think it's wise to rely on other hooks.

reactive/kubernetes_master.py
```diff
@@ -893,6 +898,10 @@ def set_final_status():
...
+        return
```
auria, line 903: The "return" statement is great to avoid calling `hookenv.status_set("active"...)`. It seems like "black magic" the need of the `@when_not` decorator on the called function, though (`missing_certificates_notice`).

auria:
LGTM. Should a unit test be created for the fixed use case (where `certificates.available` is missing)?

reactive/kubernetes_master.py
mkalcok: Just for the sake of complete discussion:

We eventually found out that moving `certificates.available` check closer to the beginning of the `set_final_status()` would set expected status immediately after CA departure. This allowed us to remove the explicit hook.

auria:
+1 from me for kubernetes-master.

tests/test_kubernetes_master.py
```diff
@@ -121,3 +121,23 @@ def test_update_certificates_with_missing_relations(mock_send_data,
...
+        assert value is not None, "Unexpected flag requested: {}".format(flag)
+        return value
+
+    mock_is_flag_set.side_effect = is_flag_set_side_effect
+
+    kubernetes_master.set_final_status()
+
+    mock_status_set.assert_called_once_with('blocked',
+                                            'Missing relation to certificate '
+                                            'authority.')
```
johnsca, line 143: These charms use [charms.unit_test](https://github.com/juju-solutions/charms.unit_test/) which, while poorly documented, [patches out the flag helpers](https://github.com/juju-solutions/charms.unit_test/blob/master/charms/unit_test.py#L269-L272) with fakes that should work the same, so this can be changed to:

```suggestion
def test_status_set_on_missing_ca():
    """Test that set_final_status() will set blocked state if CA is missing"""

    set_flag("certificates.available")
    set_flag("kubernetes-master.secure-storage.failed")

    kubernetes_master.set_final_status()
    hookenv.status_set.assert_called_with('blocked',
                                          'Failed to configure encryption; '
                                          'secrets are unencrypted or inaccessible')

    clear_flag("certificates.available")
    kubernetes_master.set_final_status()
    hookenv.status_set.assert_called_with('blocked',
                                          'Missing relation to certificate '
                                          'authority.')
```

Arguable if that's any clearer or not, but it would be a bit more consistent with the other tests. Mostly I'm just putting it out there and am +1 to merging as-is.

johnsca (approved):

PR #137, opened 2021-01-12
fixes failing execs

Comrade88:
they fail because of RBAC DENY: user "system:kubelet" groups
["system:authenticated"] cannot "create" resource "nodes/proxy".
this the same problem the role binding from rbac-proxy.yaml serves
for.
let's add another principal to the list in the binding, this fixes
the problem.


Andrey-mp:
@johnsca @Cynerva @kwmonroe Can you please check this?

Cynerva:
@Comrade88 @Andrey-mp can you provide more details on the bug? It's hard to know if this is a good solution when we don't really know what the bug is.

> they fail because of RBAC DENY: user "system:kubelet" groups
["system:authenticated"] cannot "create" resource "nodes/proxy"

Where do you see that? What are you trying to do?

Cynerva:
If you can, please file a bug about this: https://bugs.launchpad.net/charmed-kubernetes. Otherwise we might lose track of this work as our workflow is based around bugs, not PRs.

Comrade88:
> @Comrade88 @Andrey-mp can you provide more details on the bug? It's hard to know if this is a good solution when we don't really know what the bug is.
> 
> > they fail because of RBAC DENY: user "system:kubelet" groups
> > ["system:authenticated"] cannot "create" resource "nodes/proxy"
> 
> Where do you see that? What are you trying to do?

after a deploy finish we need to perform an exec inside a pod and it fails with the diagnostic above

Cynerva:
> after a deploy finish we need to perform an exec inside a pod and it fails with the diagnostic above

What are you deploying? Can you please share output of:

1. `juju status`
2. `juju config kubernetes-master authorization-mode`

What kubeconfig are you using to run the `kubectl exec` command? Is it the one from kubernetes-master/0:config ?

In normal usage we expect this to auth as system:kube-apiserver. It's weird that its authing as system:kubelet instead. We need to figure out why.

Andrey-mp:
Hi @Cynerva , I've created a bug report https://bugs.launchpad.net/charm-kubernetes-master/+bug/1916009 and added status there.
we have "authorization-mode: Node,RBAC" and "enable-keystone-authorization: true"
we are using config from .kube/config

johnsca:
From the bug, the output of `juju status`: https://pastebin.ubuntu.com/p/Xhcw8GBZWV/

From that, it appears this auth error is coming because the kubernetes-master and kubernetes-worker charms are co-located. This is not supported because both charms try to use the same client cert file, whence the username is derived, thus leading to the conflict.  A workaround may be to run the master inside a lxd container on the machine, like some of the other services are in that hyper-converged deployment.

PR #138, opened 2021-01-14
ensure secret names are unique and rfc1123 compliant

kwmonroe:
Add a random string to the generated secret names to ensure they are unique for a cluster.  Names also need to be rfc1123 compliant:

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

Fixes [LP 1911445](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1911445)

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -63,7 +63,9 @@ user-create:
   params:
     name:
       type: string
-      description: Username for the new user
+      description: |
+        Username for the new user. This value must only contain alphanumeric
+        characters, '@', '-' or '.'.
       minLength: 2
     groups:
       type: string
--- a/actions/user_actions.py
+++ b/actions/user_actions.py
@@ -66,7 +66,9 @@ def user_create():
     # TODO: make the token format less magical so it doesn't get out of
     # sync with the function that creates secrets in k8s-master.py.
     token = '{}::{}'.format(user, layer.kubernetes_master.token_generator())
-    layer.kubernetes_master.create_secret(token, user, user, groups)
+    if not layer.kubernetes_master.create_secret(token, user, user, groups):
+        action_fail('Failed to create secret for: {}'.format(user))
+        return
 
     # Create a kubeconfig
     ca_crt = layer.kubernetes_common.ca_crt_path
--- a/actions/user_actions.py
+++ b/actions/user_actions.py
@@ -1,6 +1,7 @@
 #!/usr/local/sbin/charm-env python3
 import json
 import os
+import re
 import sys
 from base64 import b64decode
 from charmhelpers.core import hookenv
@@ -60,6 +61,12 @@ def user_create():
     users = user_list()
     if user in list(users):
         action_fail('User "{}" already exists.'.format(user))
+        return
+
+    # Validate the name
+    if re.search('[^0-9A-Za-z@.-]+', user):
+        msg = "User name may only contain alphanumeric characters, '@', '-' or '.'"
+        action_fail(msg)
         return
 
     # Create the secret
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -247,8 +247,8 @@ def create_known_token(token, username, user, groups=None):
 
 def create_secret(token, username, user, groups=None):
     # secret IDs must be unique and rfc1123 compliant
-    uniq_name = re.sub('[^0-9a-z.-]+', '-', user.lower())
-    secret_id = '{}-{}-{}'.format(uniq_name, generate_rfc1123(10), AUTH_SECRET_SUFFIX)
+    sani_name = re.sub('[^0-9a-z.-]+', '-', user.lower())
+    secret_id = '{}-{}-{}'.format(sani_name, generate_rfc1123(10), AUTH_SECRET_SUFFIX)
     # The authenticator expects tokens to be in the form user::token
     token_delim = '::'
     if token_delim not in token:
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -247,8 +247,8 @@ def create_known_token(token, username, user, groups=None):
 
 def create_secret(token, username, user, groups=None):
     # secret IDs must be unique and rfc1123 compliant
-    uniq_name = re.sub('[^0-9a-z.-]+', '-', user.lower()) + generate_rfc1123(10)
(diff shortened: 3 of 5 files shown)


kwmonroe:
As noted in the linked bug, admins could create a user with non-rfc1123 data (like an uppercase letter).  The charm would use that as part of the k8s secret name, which would cause an error in secret creation:

```bash
$ juju run-action --wait kubernetes-master/0 user-create name=Bob
unit-kubernetes-master-0:
  UnitId: kubernetes-master/0
  id: "31"
  results:
    Stderr: |
      The Secret "Bob-token-auth" is invalid: metadata.name: Invalid value: "Bob-token-auth": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
    Stdout: |
      Cluster "juju-cluster" set.
      Property "users" unset.
      User "Bob" set.
      Context "juju-context" modified.
      Switched to context "juju-context".
    kubeconfig: juju scp kubernetes-master/0:/home/ubuntu/Bob-kubeconfig .
    msg: User "Bob" created.
    users: admin, system:kube-controller-manager, system:kube-proxy, system:node:juju-2a8f8a-10,
      system:node:juju-2a8f8a-11, system:node:juju-2a8f8a-12, system:kube-scheduler,
      system:monitoring, Bob
  status: completed

$ kubectl get secrets -n kube-system --kubeconfig config | grep -i bob || echo "not found"
not found
```

With this fix, the charm will ensure the secret name is valid so it can be created:
```bash
$ juju run-action --wait kubernetes-master/0 user-create name=Bob
unit-kubernetes-master-0:
  UnitId: kubernetes-master/0
  id: "33"
  results:
    Stdout: |
      Cluster "juju-cluster" set.
      Property "users" unset.
      User "Bob" set.
      Context "juju-context" modified.
      Switched to context "juju-context".
    kubeconfig: juju scp kubernetes-master/0:/home/ubuntu/Bob-kubeconfig .
    msg: User "Bob" created.
    users: admin, system:kube-controller-manager, system:kube-proxy, system:node:juju-2a8f8a-10,
      system:node:juju-2a8f8a-7, system:node:juju-2a8f8a-11, system:node:juju-2a8f8a-8,
      system:node:juju-2a8f8a-12, system:node:juju-2a8f8a-9, system:kube-scheduler,
      system:monitoring, Bob
  status: completed

$ kubectl get secrets -n kube-system --kubeconfig config | grep -i bob || echo "not found"
bob-l1xrzi1bv7-token-auth                        juju.is/token-auth                    4      95s
```

kwmonroe:
Initially, i didn't bother to validate the `name` parameter for the `user_create` action since `create_secret` now ensures valid secret names in the charm. However, if someone added a name that contained our secret delimiter (`::`), it would cause problems.

I can't think of a good use case to allow anything but alphanum, '-', and '.' for the `name` parameter, so the latest commit enforces that.

actions.yaml
```diff
@@ -63,7 +63,9 @@ user-create:
   params:
     name:
       type: string
-      description: Username for the new user
+      description: |
+        Username for the new user. This value must only contain alphanumeric
+        characters, '@', '-' or '.'.
```
Cynerva, line 68: From https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names they say alphanumeric, `-` or `.` are accepted. Why is `@` accepted here?

Likewise, those docs say only *lowercase* alphanumeric characters are valid. Is it safe for us to accept uppercase?

Tell you what, I'll give this a test deploy in a bit and see if I can answer that for myself.

Cynerva: OK, I understand now. Users aren't subject to the same restrictions as resources in Kubernetes. We accept `@` here because some people might want to use their email address as a user name. Likewise, uppercase characters are fine as part of user names, and you were careful to handle that with a .lower() when creating secrets.

Cynerva, line 68: Should `:` be allowed, too? I ask because `:` is a prominent character used in a lot of user names, as seen from `user-list`:

```
$ juju run-action --wait kubernetes-master/0 user-list
unit-kubernetes-master-0:
  UnitId: kubernetes-master/0
  id: "16"
  results:
    users: admin, Bob, bob@foo, bob, system:kube-controller-manager, system:kube-proxy,
      system:node:ip-172-31-47-88, system:node:ip-172-31-8-169, system:node:ip-172-31-21-142,
      system:kube-scheduler, system:monitoring
  status: completed
  timing:
    completed: 2021-01-15 20:08:43 +0000 UTC
    enqueued: 2021-01-15 20:08:41 +0000 UTC
    started: 2021-01-15 20:08:42 +0000 UTC
```

which, as a dumb user, makes me think I might be able to create a user named `custom:bob`, but no:

```
$ juju run-action --wait kubernetes-master/0 user-create name=custom:bob
unit-kubernetes-master-0:
  UnitId: kubernetes-master/0
  id: "20"
  message: User name may only contain alphanumeric characters, '@', '-' or '.'
  results:
    users: admin, Bob, bob@foo, bob, system:kube-controller-manager, system:kube-proxy,
      system:node:ip-172-31-47-88, system:node:ip-172-31-8-169, system:node:ip-172-31-21-142,
      system:kube-scheduler, system:monitoring
  status: failed
  timing:
    completed: 2021-01-15 20:12:14 +0000 UTC
    enqueued: 2021-01-15 20:12:08 +0000 UTC
    started: 2021-01-15 20:12:13 +0000 UTC
```

Cynerva: I think it could work to allow `:` if we updated [auth-webhook's delimiter handling](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/bb3328c490f2e05b20010f688be55296623fb7e1/templates/cdk.master.auth-webhook.py#L148) to split on only the right-most occurrence of `::`.

I don't think this is a show stopper though.

Cynerva:
This LGTM. But I would be happy to see some consideration given to the mixed messages coming from user-list and user-create regarding the `:` character.

kwmonroe:
> This LGTM. But I would be happy to see some consideration given to the mixed messages coming from user-list and user-create regarding the `:` character.

Thanks for the feedback.  I'm moving `user-list` around a bit in [lp 1906732](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1906732), so i'll tweak the action wording and/or double check the allowance of `:` in the pr for that.

PR #139, opened 2021-01-20
[kubernetes-master] allow configuration of pod security policy

niedbalski:
Allow to configure a default pod security policy, if provided
that one will overwrite the default template, if not the default
policy template will be used.

LP: #1908753

Signed-off-by: Jorge Niedbalski <jorge.niedbalski@canonical.com>


niedbalski:
@Cynerva for your review as we talk about LP: 1908753

reactive/kubernetes_master.py
```diff
@@ -1737,12 +1737,18 @@ def switch_auth_mode(forced=False):
...
-    render(
-        'rbac-pod-security-policy.yaml',
-        pod_security_policy_path,
-        {}
-    )
+    pod_security_policy = hookenv.config('pod-security-policy')
+    if pod_security_policy:
+        hookenv.log('Using configuration defined on pod-security-policy option')
+        write_file_with_autogenerated_header(pod_security_policy_path,
+                                             pod_security_policy)
```
chris-sanders, line 1744: This is user input, I think some form of validation of the input should be done. Someone could get a lot of strange errors out of this if they poorly format it.

config.yaml
```diff
@@ -343,3 +343,9 @@ options:
       respond with appropriate authentication details. For more info, please refer
       to the upstream documentation at
       https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication
+  pod-security-policy:
+    type: string
+    default: ""
+    description: |
+      Default RBAC pod security policy and privileged cluster roles.
+      https://kubernetes.io/docs/concepts/policy/pod-security-policy/
```
chris-sanders, line 351: With strings used for a lot of different things, you should probably clarify what the expected input is. Do you expect 'yaml file as string', or base64 encoded file, or something else?

reactive/kubernetes_master.py
```diff
@@ -2971,6 +2977,11 @@ def _write_encryption_config():
     )
 
 
+@when_any('config.changed.pod-security-policy')
+def pod_security_policy_config_changed():
+    clear_flag('kubernetes-master.pod-security-policy.applied')
```
chris-sanders, line 2982: I know, we're missing tests, that's in progress.
Have you tested this on a fresh deploy and not just an upgrade? I know config-changed triggers during install which can cause strange behavior. I'm not clear if clear_flag fails if the flag has never been set, it might be fine.

chris-sanders (changes requested):
A few in-line comments and questions.

reactive/kubernetes_master.py
niedbalski: There is no need for adding complexity in the validation of the policy, the policy is applied in line 1754 and any format/string error will be raised to the hookenv error log.

reactive/kubernetes_master.py
niedbalski: Yes, tested on fresh new and upgrade. Both clear the flag kubernetes-master.pod-security-policy.applied and the policy gets re-applied.

niedbalski:
@chris-sanders addressed your comments. Thanks for reviewing.

reactive/kubernetes_master.py
chris-sanders: I see what you're saying, but the use case has changed. Originally, the content was controlled by the charm and on a failure it was logged in the debug-log and retried. However, now that you're applying user input you need to surface a bad input to the user not just log it in the debug-log.

A user with a bad config would get log messages that it failed, but the charm itself isn't going to show any failures in `juju status` at least none that I see. Especially for something like a security policy I think the charm needs to provide feedback to the user since you're now accepting user input. Debug logging is fine for developers, which is where the default policy comes from.

chris-sanders:
@niedbalski I added a reply to the validation thread, that's my last standing concern.

reactive/kubernetes_master.py
niedbalski: do you have any suggestion on how to implement such validation?

adam-stokes:
Hi @niedbalski we'd also like to see some sort of test added to this PR. You can find an example of our integration tests at https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/tests/integration/test_kubernetes_master_integration.py and you could even add to the test suite in that same file. It uses pytest-operator and libjuju under the hood. The test already covers building and deploying a kubernetes-core bundle with the local charm built. So you can start with that assumption that a deployment would be available to your tests

adam-stokes (changes requested):
Hi @niedbalski we'd also like to see some sort of test added to this PR. You can find an example of our integration tests at https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/tests/integration/test_kubernetes_master_integration.py and you could even add to the test suite in that same file. It uses pytest-operator and libjuju under the hood. The test already covers building and deploying a kubernetes-core bundle with the local charm built. So you can start with that assumption that a deployment would be available to your tests

niedbalski:
@here added functional tests as required.

niedbalski:
I don't think the test failure is related to the patch.

niedbalski:
@Cynerva hey George, mind to review this patch? Thanks!

Cynerva:
@niedbalski Thanks for the ping, somehow this PR was missing from our team's "to review" column in Trello. It's on there now. Sorry for the delay.

I have some other reviews to get through but will hopefully get to this one today or tomorrow.

tests/integration/test_kubernetes_master_integration.py
```diff
@@ -15,6 +15,16 @@ async def test_build_and_deploy(ops_test):
...
 
 
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy_pod_security_policy(ops_test):
+    bundle = ops_test.render_bundle(
+        "tests/data/bundle_pod_security_policy.yaml",
+        master_charm=await ops_test.build_charm("."),
+    )
+    await ops_test.model.deploy(bundle)
+    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=60 * 60)
```
Cynerva, line 25: This test is failing:

```
__________________ test_build_and_deploy_pod_security_policy ___________________
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/tests/integration/test_kubernetes_master_integration.py", line 24, in test_build_and_deploy_pod_security_policy
    await ops_test.model.deploy(bundle)
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1433, in deploy
    await handler.execute_plan()
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/bundle.py", line 167, in execute_plan
    self.references[step.id_] = await change.run(self)
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/bundle.py", line 346, in run
    await context.model._deploy(
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1570, in _deploy
    result = await app_facade.Deploy(applications=[app])
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 480, in wrapper
    reply = await f(*args, **kwargs)
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/client/_client8.py", line 784, in Deploy
    reply = await self.rpc(msg)
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 623, in rpc
    result = await self.connection.rpc(msg, encoder=TypeEncoder)
  File "/home/ubuntu/actions-runner/_work/charm-kubernetes-master/charm-kubernetes-master/.tox/integration/lib/python3.8/site-packages/juju/client/connection.py", line 493, in rpc
    raise errors.JujuError(err_results)
juju.errors.JujuError: ['cannot add application "containerd": application already exists']
```

The tests in this module are run in sequence, and they share the same Juju model. So by the time `test_build_and_deploy_security_policy` runs, `test_build_and_deploy` has already deployed the charms into the model.

I think to test this properly, you will need to write a test_pod_security_policy that assumes the charms have already been deployed. I would expect the test to set the kubernetes-master application's pod-security-policy config, and then verify that the pod security policy has been updated in Kubernetes.

Cynerva (changes requested):

tests/integration/test_kubernetes_master_integration.py
niedbalski: OK, I am taking this approach and will update this PR.

niedbalski:
@Cynerva hey george, all tests passing. can you please review/merge this proposal? thanks!

tests/data/bundle_pod_security_policy.yaml
```diff
@@ -0,0 +1,145 @@
...
+- - flannel:etcd
+  - etcd:db
+- - flannel:cni
+  - kubernetes-master:cni
+- - flannel:cni
+  - kubernetes-worker:cni
+- - containerd:containerd
+  - kubernetes-worker:container-runtime
+- - containerd:containerd
+  - kubernetes-master:container-runtime
```
Cynerva, line 145: It's not sustainable for us, as maintainers, to create a new bundle for every config option we need to test.

There are 31 config options in kubernetes-master. Can you imagine if we had 31 bundles in here, but needed to change something like a constraint, or any architectural changes to the relations? We would have to update all 31 bundles and make sure none get missed.

tests/integration/test_kubernetes_master_integration.py
```diff
@@ -7,9 +7,10 @@
 
 
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy(ops_test):
+async def test_build_and_deploy_pod_security_policy(ops_test):
     bundle = ops_test.render_bundle(
-        "tests/data/bundle.yaml", master_charm=await ops_test.build_charm(".")
+        "tests/data/bundle_pod_security_policy.yaml",
+        master_charm=await ops_test.build_charm("."),
```
Cynerva, line 13: It looks like you overwrote test_build_and_deploy here, why?

tests/integration/test_kubernetes_master_pod_security.py
```diff
@@ -0,0 +1,27 @@
...
+
+async def test_status_messages(ops_test):
+    """ Validate that the status messages are correct. """
+    expected_messages = {
+        "kubernetes-master": "Kubernetes master running.",
+        "kubernetes-worker": "Kubernetes worker running.",
+    }
+    for app, message in expected_messages.items():
+        for unit in ops_test.model.applications[app].units:
+            assert unit.workload_status_message == message
```
Cynerva, line 27: This file is almost entirely a duplicate of test_kubernetes_master_integration.py

Cynerva (changes requested):
Sorry, the integration tests in this PR need more attention. There are three major problems with them as they are now:

1. There is a lot of code duplication that will make these tests hard to maintain.

If we merged this, we would essentially have two copies of the bundle and two copies of test_kubernetes_master_integration with only slight differences between them. We can't take that approach for every test that we write, or the tests would quickly become unmaintainable.

2. The tests will take a long time to run because they deploy from scratch every time.

It would be nice if every test could run from a clean slate, but the reality is that deployments take a very long time and we will not be able to run tests in a timely manner if we redeploy between each test.

3. The tests don't verify that the pod-security-policy is actually written to Kubernetes.

We need tests to verify that the config option is actually doing what it's meant to.

---

Again, I think what you need to do is leave test_build_and_deploy alone, don't add a new bundle or new test file, and instead update test_kubernetes_master_integration.py so it looks something like:

```
...

@pytest.mark.abort_on_fail
async def test_build_and_deploy(ops_test):
    ...

async def test_status_messages(ops_test):
    ...

async def test_pod_security_policy(ops_test):
    # Set pod-security-policy config on kubernetes-master
    # Verify pod-security-policy has been updated in Kubernetes
    # Reset pod-security-policy config to empty
    # Verify pod-security-policy has been reverted in Kubernetes
```

You can access the Juju model via `ops_test.model`, and examples of how to use the model object can be found in the [python-libjuju repo](https://github.com/juju/python-libjuju).

Hopefully that makes sense. If you have any questions, let us know.

johnsca:
Following up on this, aside from the conflicts to be resolved, I would also like to see the test refactored to just be another case added to the existing `tests/integration/test_kubernetes_master_integration.py` which uses `ops_test.model.applications["kubernetes-master"].set_config({...})` to update the new config value on the existing deployment and testing from there.  In fact, there's now an example of a test [doing just that](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/4ade8a1/tests/integration/test_kubernetes_master_integration.py#L94-L96) in that file that you can follow.

Cynerva:
New PR with updated tests: https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/178

Closing this one. Thanks folks.

PR #142, opened 2021-02-02
reuse existing secret names and make secret UIDs more useful

kwmonroe:
If a secret exists for a user/service, reuse that name to ensure users/services don't have multiple secrets at once. This is needed since #138 introduced random strings in secret names to make them unique, hence we can no longer easily construct a secret name based on the user/service alone.

With reusable unique secret names, we are free to make the uid field more accurate -- instead of `kubelet-0`, we'll use `kubernetes-worker-0`.  This makes it easy for an admin to spot which secret belongs to which unit, and becomes especially useful when, for example, `k8s-worker` is deployed multiple times with different application names.

While we're here, allow `:` in usernames when creating new users. We have services like `system:node:foo` that show up in our user list, but had previously not allowed `:` when running the `user-create` action. This could be confusing, and there's no reason to disallow that character, so stop doing that.

Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1906732

### DIFF
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -325,6 +325,7 @@ def get_secret_names():
     except (CalledProcessError, FileNotFoundError):
         # The api server may not be up, or we may be trying to run kubelet before
         # the snap is installed. Send back an empty dict.
+        hookenv.log('Unable to get existing secrets', level=hookenv.WARNING)
         return {}
 
     secrets = json.loads(output)
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -254,9 +254,15 @@ def create_known_token(token, username, user, groups=None):
 
 
 def create_secret(token, username, user, groups=None):
-    # secret IDs must be unique and rfc1123 compliant
-    sani_name = re.sub('[^0-9a-z.-]+', '-', user.lower())
-    secret_id = 'auth-{}-{}'.format(sani_name, generate_rfc1123(10))
+    secrets = get_secret_names()
+    if username in secrets:
+        # Use existing secret ID if one exists for our username
+        secret_id = secrets[username]
+    else:
+        # secret IDs must be unique and rfc1123 compliant
+        sani_name = re.sub('[^0-9a-z.-]+', '-', user.lower())
+        secret_id = 'auth-{}-{}'.format(sani_name, generate_rfc1123(10))
+
     # The authenticator expects tokens to be in the form user::token
     token_delim = '::'
     if token_delim not in token:
@@ -307,6 +313,32 @@ def get_csv_password(csv_fname, user):
                 # probably a blank line or comment; move on
                 continue
     return None
+
+
+def get_secret_names():
+    '''Return a dict of 'username: secret_id' for Charmed Kubernetes users.'''
+    try:
+        output = kubernetes_common.kubectl(
+            'get', 'secrets', '-n', AUTH_SECRET_NS,
+            '--field-selector', 'type={}'.format(AUTH_SECRET_TYPE),
+            '-o', 'json').decode('UTF-8')
+    except (CalledProcessError, FileNotFoundError):
+        # The api server may not be up, or we may be trying to run kubelet before
+        # the snap is installed. Send back an empty dict.
+        return {}
+
+    secrets = json.loads(output)
+    secret_names = {}
+    if 'items' in secrets:
+        for secret in secrets['items']:
+            try:
+                secret_id = secret['metadata']['name']
+                username_b64 = secret['data']['username'].encode('UTF-8')
+            except (KeyError, TypeError):
+                # CK secrets will have populated 'data', but not all secrets do
+                continue
+            secret_names[b64decode(username_b64).decode('UTF-8')] = secret_id
+    return secret_names
 
 
 def get_secret_password(username):
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -1289,8 +1289,10 @@ def create_tokens_and_sign_auth_requests():
             continue
         kubelet_token = get_token(username)
         if not kubelet_token:
-            # Usernames have to be in the form of system:node:<nodeName>
-            userid = "kubelet-{}".format(request[0].split('/')[1])
+            # Username will be in the form of system:node:<nodeName>.
+            # User ID will be a worker <unitName>, and while not used today, we store
+            # this in case it becomes useful to map a secret to a unit in the future.
+            userid = request[0]
             setup_tokens(None, username, userid, group)
             kubelet_token = get_token(username)
         # When bootstrapping a new cluster, we may not have all our secrets yet.
--- a/templates/cdk.master.auth-webhook.py
+++ b/templates/cdk.master.auth-webhook.py
@@ -145,7 +145,7 @@ def check_secrets(token_review):
                 username = uid = b64decode(username_b64).decode('UTF-8')
                 pw_delim = '::'
                 if pw_delim in password:
-                    uid = password.split(pw_delim)[0]
+                    uid = password.rsplit(pw_delim, 1)[0]
                 groups = b64decode(groups_b64).decode('UTF-8').split(',')
                 token_review['status'] = {
                     'authenticated': True,
--- a/actions.yaml
+++ b/actions.yaml
@@ -65,7 +65,7 @@ user-create:
       type: string
       description: |
         Username for the new user. This value must only contain alphanumeric
-        characters, '@', '-' or '.'.
+        characters, ':', '@', '-' or '.'.
       minLength: 2
     groups:
       type: string
--- a/actions/user_actions.py
+++ b/actions/user_actions.py
@@ -64,8 +64,8 @@ def user_create():
         return
 
     # Validate the name
-    if re.search('[^0-9A-Za-z@.-]+', user):
-        msg = "User name may only contain alphanumeric characters, '@', '-' or '.'"
+    if re.search('[^0-9A-Za-z:@.-]+', user):
+        msg = "User name may only contain alphanumeric characters, ':', '@', '-' or '.'"
         action_fail(msg)
         return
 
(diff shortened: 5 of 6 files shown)


kwmonroe:
## Test fresh deployment

Secrets created *with* the new naming format:
```bash
$ kubectl get secrets -n kube-system --kubeconfig config|grep juju
auth-admin-y1jdwai.hu                            juju.is/token-auth                    4      19m
auth-kube-controller-manager-svvw5aa1ip          juju.is/token-auth                    4      19m
auth-kube-proxy-10xms53482                       juju.is/token-auth                    4      19m
auth-kubernetes-worker-0-p8v1a6l1bt              juju.is/token-auth                    4      19m
auth-kubernetes-worker-1-ieq1f8a0q5              juju.is/token-auth                    4      16m
auth-kubernetes-worker-2-ivr82v4i0h              juju.is/token-auth                    4      12m
auth-system-kube-scheduler-d6km85o6f3            juju.is/token-auth                    4      19m
auth-system-monitoring-fjbv0s1vfw                juju.is/token-auth                    4      19m
```

Test admin password change, which will trigger `authentication.setup` handlers. Pre-change (note unique secret name):
```bash
$ kubectl get secret -n kube-system auth-admin-y1jdwai.hu --kubeconfig config -o yaml
...
  password: YWRtaW46OkpaN29yODlDS1RRNWYyMzB0V1AwaGVBcUo4VGFCell5
...
  name: auth-admin-y1jdwai.hu
```

Post-change (same unique admin secret name has been updated with new password):
```
$ juju config kubernetes-master client_password=passw0rd
$ kubectl get secret -n kube-system auth-admin-y1jdwai.hu --kubeconfig config -o yaml
...
  password: YWRtaW46OnBhc3N3MHJk
...
  name: auth-admin-y1jdwai.hu
```

All secrets remain as we expect:
```bash
$ kubectl get secrets -n kube-system --kubeconfig config|grep juju
auth-admin-y1jdwai.hu                            juju.is/token-auth                    4      80m
auth-kube-controller-manager-svvw5aa1ip          juju.is/token-auth                    4      80m
auth-kube-proxy-10xms53482                       juju.is/token-auth                    4      80m
auth-kubernetes-worker-0-p8v1a6l1bt              juju.is/token-auth                    4      79m
auth-kubernetes-worker-1-ieq1f8a0q5              juju.is/token-auth                    4      76m
auth-kubernetes-worker-2-ivr82v4i0h              juju.is/token-auth                    4      73m
auth-system-kube-scheduler-d6km85o6f3            juju.is/token-auth                    4      80m
auth-system-monitoring-fjbv0s1vfw                juju.is/token-auth                    4      79m
```

## Test upgrade from 1.20

Secrets created *without* the new naming format:
```bash
$ kubectl get secrets -n kube-system --kubeconfig config|grep juju
admin-token-auth                                 juju.is/token-auth                    4      5m39s
kube-controller-manager-token-auth               juju.is/token-auth                    4      5m38s
kube-proxy-token-auth                            juju.is/token-auth                    4      5m38s
kubelet-0-token-auth                             juju.is/token-auth                    4      5m36s
kubelet-1-token-auth                             juju.is/token-auth                    4      3m58s
kubelet-2-token-auth                             juju.is/token-auth                    4      2m21s
system-kube-scheduler-token-auth                 juju.is/token-auth                    4      5m37s
system-monitoring-token-auth                     juju.is/token-auth                    4      5m37s
```

Upgrade k8s-master with this fix:
```bash
$ juju upgrade-charm kubernetes-master --path ~/charms/builds/kubernetes-master
```

Test admin password change, which will trigger `authentication.setup` handlers. Pre-change (note non-unique secret name):
```bash
$ kubectl get secret -n kube-system admin-token-auth --kubeconfig config -o yaml
...
  password: YWRtaW46Omd2WGkzb0l1UDd6MzA4bVdLZ1V4eTQxNEkzaVJBTWFy
...
  name: admin-token-auth
```

Post-change (same non-unique admin secret name has been updated with new password):
```bash
$ juju config kubernetes-master client_password=passw0rd
$ kubectl get secret -n kube-system admin-token-auth --kubeconfig config -o yaml
...
  password: YWRtaW46OnBhc3N3MHJk
...
  name: admin-token-auth
```

Add a unit to see that the new secret follows the new format, and existing secrets remain as we expect:
```bash
$ juju add-unit kubernetes-worker
$ kubectl get secrets -n kube-system --kubeconfig config|grep juju
admin-token-auth                                 juju.is/token-auth                    4      31m
auth-kubernetes-worker-3-bm6un.lucg              juju.is/token-auth                    4      6m14s
kube-controller-manager-token-auth               juju.is/token-auth                    4      31m
kube-proxy-token-auth                            juju.is/token-auth                    4      31m
kubelet-0-token-auth                             juju.is/token-auth                    4      31m
kubelet-1-token-auth                             juju.is/token-auth                    4      29m
kubelet-2-token-auth                             juju.is/token-auth                    4      27m
system-kube-scheduler-token-auth                 juju.is/token-auth                    4      31m
system-monitoring-token-auth                     juju.is/token-auth                    4      31m
```

reactive/kubernetes_master.py
```diff
@@ -1289,8 +1289,10 @@ def create_tokens_and_sign_auth_requests():
             continue
         kubelet_token = get_token(username)
         if not kubelet_token:
-            # Usernames have to be in the form of system:node:<nodeName>
-            userid = "kubelet-{}".format(request[0].split('/')[1])
+            # Username will be in the form of system:node:<nodeName>.
+            # User ID will be a worker <unitName>, and while not used today, we store
+            # this in case it becomes useful to map a secret to a unit in the future.
+            userid = request[0]
```
Cynerva, line 1295: I'm starting to think we could save a lot of confusion if we just use the username as the uid in all cases, i.e. instead of

```
username: system:node:ip-172-31-37-122
uid: kubernetes-worker/0
```

we use the same for both:

```
username: system:node:ip-172-31-37-122
uid: system:node:ip-172-31-37-122
```

templates/cdk.master.auth-webhook.py
```diff
@@ -145,7 +145,7 @@ def check_secrets(token_review):
                 username = uid = b64decode(username_b64).decode('UTF-8')
                 pw_delim = '::'
                 if pw_delim in password:
-                    uid = password.split(pw_delim)[0]
+                    uid = password.rsplit(pw_delim, 1)[0]
```
Cynerva, line 148: Do we need this delimiter splitting at all? It looks like uid is already stored as a separate field in the secret. Why don't we get it from there?

lib/charms/layer/kubernetes_master.py
```diff
@@ -309,6 +315,32 @@ def get_csv_password(csv_fname, user):
...
+    '''Return a dict of 'username: secret_id' for Charmed Kubernetes users.'''
+    try:
+        output = kubernetes_common.kubectl(
+            'get', 'secrets', '-n', AUTH_SECRET_NS,
+            '--field-selector', 'type={}'.format(AUTH_SECRET_TYPE),
+            '-o', 'json').decode('UTF-8')
+    except (CalledProcessError, FileNotFoundError):
+        # The api server may not be up, or we may be trying to run kubelet before
+        # the snap is installed. Send back an empty dict.
+        return {}
```
Cynerva, line 328: I would at least log a warning here

Cynerva:
LGTM. I had a few comments but no show stoppers, just things to think about I guess.

reactive/kubernetes_master.py
Cynerva: I mention this because `uid: kubernetes-worker/0` raises questions:

1. Is the `/` safe?
2. What happens if we relate to a kubernetes-worker application in a different model? Will we end up with multiple users that have the same `kubernetes-worker/0` UID?

lib/charms/layer/kubernetes_master.py
Cynerva: If the kubectl call here fails and we return `{}`, then create_secret will move on as if the secret doesn't exist, and try to create a new secret. We could end up with duplicate secrets in that case right?

reactive/kubernetes_master.py
kwmonroe: It would potentially be less confusing to make uid==username, but we use the uid in kubeconfig tokens; it's nice to me to see the unit name in there so i know at-a-glance that a kubeconfig goes with a $unit vs a $hostname.

> 1. Is the `/` safe?

Looks safe. We expose the `$unit` as part of a token when validating secrets, and that works fine:

```bash
$ curl -X POST -H "Content-Type: application/json" -d '{"kind": "TokenReview", "apiVersion": "authentication.k8s.io/v1beta1", "spec": {"token": "worker-a/3::r0DzvhNmcWqDLtP72l8Ux2ZJNm5DvKJp"}}' https://10.246.153.172:5000/v1beta1
{"apiVersion":"authentication.k8s.io/v1beta1","kind":"TokenReview","spec":{"token":"worker-a/3::r0DzvhNmcWqDLtP72l8Ux2ZJNm5DvKJp"},"status":{"authenticated":true,"user":{"groups":["system:nodes"],"uid":"worker-a/3","username":"system:node:juju-2a8f8a-15"}}}
```

> 2. What happens if we relate to a kubernetes-worker application in a different model? Will we end up with multiple users that have the same `kubernetes-worker/0` UID?

Yes, we'll have multiple users with the same uid, but they will have unique secret names, and the `auth-webhook` doesn't care about uid, so it'll keep looping through all secrets until it finds a match.

templates/cdk.master.auth-webhook.py
kwmonroe: > Do we need this delimiter splitting at all? It looks like uid is already stored as a separate field in the secret. Why don't we get it from there?

I like it as a separator in kubeconfig files, so i see:
```
token: worker-a/3::r0DzvhNmcWqDLtP72l8Ux2ZJNm5DvKJp
```
instead of:
```
token: worker-a/3r0DzvhNmcWqDLtP72l8Ux2ZJNm5DvKJp
```

lib/charms/layer/kubernetes_master.py
kwmonroe: > If the kubectl call here fails and we return `{}`, then create_secret will move on as if the secret doesn't exist, and try to create a new secret. We could end up with duplicate secrets in that case right?

That's true, but if the apiserver is down for `get_secret_names`, then it will *probably* still be down for the subsequent `kubectl apply`.  I plan to refactor the handling of a "down" apiserver in 1.21.

PR #143, opened 2021-02-04
add kubeconfig to kubectl commands in actions

hemanthnakkina:
Following actions namespace-create, namespace-list, namespace-delete
run kubectl command without kubeconfig parameter. This is required
since the HOME environment variable is not set, so no default
kubeconfig file is picked.

Add -kubeconfig=/home/ubuntu/config to kubectl command

Closes-Bug: #1914359

### DIFF
--- a/actions/namespace-create
+++ b/actions/namespace-create
@@ -15,7 +15,8 @@ os.environ['PATH'] += os.pathsep + os.path.join(os.sep, 'snap', 'bin')
 
 
 def kubectl(args):
-    cmd = ['kubectl'] + args
+    cmd = ["kubectl", "--kubeconfig=/home/ubuntu/config"]
+    cmd.extend(args)
     return check_output(cmd)
 
 


actions/namespace-create
```diff
@@ -15,7 +15,8 @@ os.environ['PATH'] += os.pathsep + os.path.join(os.sep, 'snap', 'bin')
 
 
 def kubectl(args):
-    cmd = ['kubectl'] + args
+    cmd = ["kubectl", "--kubeconfig=/home/ubuntu/config"]
```
pponnuvel, line 18: Question: would it be better to use `$HOME` instead of `/home/ubuntu`?

hemanthnakkina: > Question: would it be better to use `$HOME` instead of `/home/ubuntu`?

Pon, $HOME is not defined for execution of actions and hence the problem. Please see [1]
I have used /home/ubuntu instead of /root/.kube just to make this consistent across other actions
  
[1] https://bugs.launchpad.net/charm-kubernetes-master/+bug/1914359/comments/1

pponnuvel: Thanks, Hemanth!

johnsca (approved):
LGTM.  I agree that, while it would've made more sense to to use `/root/.kube/config` for the actions since that's generally intended to be what the charm code uses, the other actions (in `kubectl-actions.py`) do currently use `/home/ubuntu/.kube/config` and either should work, so it's fine to keep it consistent.  If we want, we can go back and update them both at a later time.

PR #144, opened 2021-02-06
Prevent race conditions for config creation

GlassOfWhiskey:
Some of the kubeconfig files created in the
build_kubeconfig() function need the related
secret to be already present in the cluster.
The creation of such secrets is managed in the
create_tokens_and_sign_auth_requests() function,
called in the setup_leader_authentication() function.

This last function needs `authentication.setup` to
be unset, but does not explicitly need the apiserver
to be configured. Nevertheless, the setup_tokens()
function does not create a secret if the apiserver
is not configured. This can lead to race conditions.

Explicitly calling again the build_kubeconfig() and the
create_tokens_and_sign_auth_requests() function in the
setup_auth_webhook_tokens(), i.e. after the apiserver
configuration, ensures that all secrets are created
and all configuration files are saved, preventing race
conditions and the consequent missing files errors.

Closes-Bug: #1914658


adam-stokes (changes requested):
Hi @GlassOfWhiskey we'd also like to see some sort of test added to this PR. You can find an example of our integration tests at https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/tests/integration/test_kubernetes_master_integration.py and you could even add to the test suite in that same file. It uses pytest-operator and libjuju under the hood. The test already covers building and deploying a kubernetes-core bundle with the local charm built. So you can start with that assumption that a deployment would be available to your tests

GlassOfWhiskey:
Hi @adam-stokes, sorry but it is not clear to me which kind of tests would you like me to add.

This PR prevents race conditions in the deployment process, by making explicit the true dependency between the `build_kubeconfig()` function and the `'kubernetes-master.apiserver.configured'` event. Being race conditions non-deterministic, it is hard to test their behaviour with an automated test. In the "right" environment, where rightness is mainly related to the deployer's luck, it is also possible that they never arise.

Empirical tests coming from my OpenStack production environment confirm the correctness of the fix: no race conditions happened with the modified version of the charm, while before about half of the deployments were failing due to missing configurations.

Moreover, the correctness of the fix can also easily be figured out with a "human" code inspection, realising how the aforementioned dependency was not explicitly coded before the fix. Conversely, the need of such a dependency is explicitly documented in the body of the `setup_tokens()` function, which delays the creation of secret `until the apiserver is configured` (therefore requiring to be called at least once after the apiserver creation).

Cynerva:
I don't think a new test is needed. [test_build_and_deploy](https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/0320324fef6c74841e7f023fee424fce4ce3d63a/tests/integration/test_kubernetes_master_integration.py#L10) should cover this. If masters become active/idle then we did not hit the race condition.

Cynerva:
LGTM, just need the tests to pass. Currently blocked by a snap store outage.

PR #149, opened 2021-03-17
Fix token creation naming and retry logic

johnsca:
Avoid the possibility of invalid runs of dashes or dots in token names by removing them from the list of characters. Also ensure that failed token creation is actually retried and report such failures in status.

Fixes [lp:1916780][]

[lp:1916780]: https://bugs.launchpad.net/charm-kubernetes-worker/+bug/1916780


kwmonroe (approved):
LGTM, but geez, what were the chances you'd get `..` or `-.` or `--` in the middle of that random string?!? Some might say > 0, and they'd be right.

Thanks for the fix!

PR #153, opened 2021-04-06
Add support for loadbalancer interface

johnsca:
This adds support for the `loadbalancer` interface so that cloud-native LBs can be provided by the integrator charms. Additionally, it simplifies the confusing way the relations between the masters and workers change depending on whether kubeapi-load-balancer is being used or not by making that use the same `lb-provider` endpoint and always forwarding the API endpoint URLs via the `kube-control` relation.

Part of [lp:1921776][]
Depends on:
  * https://github.com/juju-solutions/loadbalancer-interface/pull/13
  * https://github.com/juju-solutions/interface-kube-control/pull/33
  * https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/84
  * https://github.com/charmed-kubernetes/charm-kubeapi-load-balancer/pull/11

[lp:1921776]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1921776

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -29,9 +29,13 @@ peers:
     interface: kube-masters
 provides:
   kube-api-endpoint:
-    # kube-api-endpoint is deprecated. Its functionality has been rolled into
-    # the kube-control interface. The relation endpoint will be removed in a
-    # future release.
+    # Use of this relation is strongly discouraged as the API endpoints will be
+    # provided via the kube-control relation. However, it can be used to
+    # override those endpoints if you need to inject a reverse proxy between
+    # the master and workers using a charm which only supports the old MITM
+    # style relations. Note, though, that since this reverse proxy will not be
+    # visible to the master, it will not be used in any of the client or
+    # component kube config files.
     interface: http
   cluster-dns:
     # kube-dns is deprecated. Its functionality has been rolled into the
@@ -54,6 +58,8 @@ requires:
   etcd:
     interface: etcd
   loadbalancer:
+    # Use of this relation is strongly discouraged in favor of the more
+    # explicit loadbalancer-internal / loadbalancer-external relations.
     interface: public-address
   ceph-storage:
     interface: ceph-admin
@@ -73,7 +79,13 @@ requires:
     interface: keystone-credentials
   dns-provider:
     interface: kube-dns
-  lb-provider:
+  loadbalancer-internal:
+    # Indicates that the LB should not be public and should use internal
+    # networks if available. Intended for control plane and other internal use.
+    interface: loadbalancer
+  loadbalancer-external:
+    # Indicates that the LB should be public facing. Intended for clients which
+    # must reach the API server via external networks.
     interface: loadbalancer
 resources:
   core:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -29,6 +29,9 @@ peers:
     interface: kube-masters
 provides:
   kube-api-endpoint:
+    # kube-api-endpoint is deprecated. Its functionality has been rolled into
+    # the kube-control interface. The relation endpoint will be removed in a
+    # future release.
     interface: http
   cluster-dns:
     # kube-dns is deprecated. Its functionality has been rolled into the
@@ -70,6 +73,8 @@ requires:
     interface: keystone-credentials
   dns-provider:
     interface: kube-dns
+  lb-provider:
+    interface: loadbalancer
 resources:
   core:
     type: file
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -863,10 +863,10 @@ def set_final_status():
         hookenv.status_set("waiting", "Waiting for API server to be configured")
         return
 
-    is_leader = is_state("leadership.is_leader")
-    authentication_setup = is_state("authentication.setup")
-    if not is_leader and not authentication_setup:
-        hookenv.status_set("waiting", "Waiting on leader's crypto keys.")
+    auth_setup = is_flag_set("authentication.setup")
+    webhook_tokens_setup = is_flag_set("kubernetes-master.auth-webhook-tokens.setup")
+    if auth_setup and not webhook_tokens_setup:
+        hookenv.status_set("waiting", "Failed to setup auth-webhook tokens; will retry")
         return
 
     if is_state("kubernetes-master.components.started"):
@@ -887,10 +887,10 @@ def set_final_status():
 
     # Note that after this point, kubernetes-master.components.started is
     # always True.
-
-    webhook_tokens_setup = is_flag_set("kubernetes-master.auth-webhook-tokens.setup")
-    if not webhook_tokens_setup:
-        hookenv.status_set("waiting", "Failed to setup auth-webhook tokens; will retry")
+    is_leader = is_state("leadership.is_leader")
+    authentication_setup = is_state("authentication.setup")
+    if not is_leader and not authentication_setup:
+        hookenv.status_set("waiting", "Waiting on leader's crypto keys.")
         return
 
     addons_configured = is_state("cdk-addons.configured")
@@ -1156,7 +1156,6 @@ def register_auth_webhook():
 
 @when(
     "kubernetes-master.apiserver.configured",
-    "kubernetes-master.components.started",
     "kubernetes-master.auth-webhook-service.started",
     "authentication.setup",
 )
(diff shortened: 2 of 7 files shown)


metadata.yaml
```diff
@@ -29,6 +29,9 @@ peers:
     interface: kube-masters
 provides:
   kube-api-endpoint:
+    # kube-api-endpoint is deprecated. Its functionality has been rolled into
+    # the kube-control interface. The relation endpoint will be removed in a
+    # future release.
```
Cynerva, line 34: kubernetes-e2e and tigera-secure-ee will also need to be updated since they depend on this endpoint:

https://github.com/charmed-kubernetes/charm-kubernetes-e2e/blob/96078da927cda0de6448fbe25ba16c5071966177/metadata.yaml#L21-L22

https://github.com/charmed-kubernetes/layer-tigera-secure-ee/blob/99f4b08d3b3f7e6efcd11aca14150c617bfdb791/metadata.yaml#L27-L28

Cynerva: I think we lose something by deprecating kube-api-endpoint. It was possible for users to stand up a reverse-proxy (e.g. apache/nginx) in front of the Kubernetes API. Do we know for sure that users aren't dependent on this?

reactive/kubernetes_master.py
```diff
@@ -1410,12 +1417,47 @@ def push_service_data():
...
 
-@when("certificates.available", "kube-api-endpoint.available", "cni.available")
+@when("leadership.is_leader")
+@when("endpoint.lb-provider.available")
+@when_not("kubernetes-master.sent-lb-request")
+def request_load_balancer():
+    """Request a LB from the related provider."""
+    lb_provider = endpoint_from_name("lb-provider")
+    req = lb_provider.get_request("api-server")
+    req.protocol = req.protocols.tcp
```
Cynerva, line 1427: Should this be https?

reactive/kubernetes_master.py
```diff
@@ -1410,12 +1417,47 @@ def push_service_data():
...
+    port = kubernetes_master.STANDARD_API_PORT
+    req.port_mapping = {port: port}
+    req.public = True
+    if not req.health_checks:
+        req.add_health_check(
+            protocol=req.protocols.http,
+            port=8080,
+            path="/livez",
```
Cynerva, line 1435: kube-apiserver is listening on https port 6443, right? What's listening on http port 8080?

reactive/kubernetes_master.py
```diff
@@ -1410,12 +1417,47 @@ def push_service_data():
...
+        )
+    lb_provider.send_request(req)
+    set_flag("kubernetes-master.sent-lb-request")
```
Cynerva, line 1438: Looks like this flag is never cleared. What if the user removes the kubernetes-master:lb-provider relaton and re-adds it? We would need to send a new request, but we never will, right?

reactive/kubernetes_master.py
johnsca: Ah, good catch.  I'm not sure where I got that from.

reactive/kubernetes_master.py
johnsca: No, because we don't need stateful HTTPS routing features and most LB providers don't support them anyway.

reactive/kubernetes_master.py
johnsca: That was a case of me double-fixing getting duplicate health checks.  The flag technically isn't even needed, so I'll just drop it.

metadata.yaml
johnsca: The relation pattern doesn't affect whether you can put a reverse proxy in front of the API server; in fact, that's exactly what the kubapi-load-balancer charm is doing.  It just changes how the information about that information is disseminated between the charms.

For comparison, let me outline the current situation vs this proposal.

## Current

The current setup depends on whether you're using kubeapi-load-balancer or not:

### Without kubeapi-load-balancer

  * k8s-master <-- kube-api-endpoint --> k8s-worker (communicates API endpoint address to workers)
  * k8s-master <-- kube-control --> k8s-worker (communicates all other control plane info to workers)
  * k8s-master <-- lb-provider --> cloud integrator, (maybe) haproxy, etc (optional)

### With kubeapi-load-balancer

  * k8s-master <-- loadbalancer --> kubeapi-load-balancer (communicates the proxy API endpoint address to the master)
  * k8s-master <-- kube-api-endpoint --> kubeapi-load-balancer (communicates the actual API endpoint addresses to the proxy)
  * k8s-master <-- kube-control --> k8s-worker (communicates all the other control plane info to the workers)
  * kubeapi-load-balancer <-- kube-api-endpoint --> k8s-worker (communicates the proxy API endpoint address to the workers)

## Proposal

For all cases it would be:

  * k8s-master <-- kube-control --> k8s-worker (communicates all control plane info, including API endpoint, to the workers)
  * k8s-master <-- lb-provider --> kubeapi-load-balancer, cloud integrator, (maybe) haproxy, etc (optional)

In this case, the kubeapi-load-balancer becomes just another optional LB provider.

Cynerva: Right, I get that. My concern is that kubeapi-load-balancer might not be the only reverse proxy people are using. Are we going to add new loadbalancer relations to charms like haproxy and apache2 as well? Just how much work are we creating by making this change?

Cynerva (changes requested):
After relating `kubernetes-master:loadbalancer-internal` and `kubernetes-master:loadbalancer-external` to `kubeapi-load-balancer:lb-consumers`:

```
unit-kubernetes-master-0: 15:54:04 ERROR unit.kubernetes-master/0.juju-log loadbalancer-internal:18: Hook error:
  File "/var/lib/juju/agents/unit-kubernetes-master-0/.venv/lib/python3.8/site-packages/charms/reactive/__init__.py", line 74, in main
    bus.dispatch(restricted=restricted_mode)
  File "/var/lib/juju/agents/unit-kubernetes-master-0/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 390, in dispatch
    _invoke(other_handlers)
  File "/var/lib/juju/agents/unit-kubernetes-master-0/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 359, in _invoke
    handler.invoke()
  File "/var/lib/juju/agents/unit-kubernetes-master-0/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 181, in invoke
    self._action(*args)
  File "/var/lib/juju/agents/unit-kubernetes-master-0/charm/reactive/kubernetes_master.py", line 2087, in build_kubeconfig
    internal_url = kubernetes_master.get_api_url(internal_endpoints)
  File "lib/charms/layer/kubernetes_master.py", line 154, in get_api_url
    return urls[kubernetes_common.get_unit_number() % len(urls)]
ZeroDivisionError: integer division or modulo by zero
```

johnsca:
[Here's](https://pastebin.ubuntu.com/p/yG6ZFBrrVY/) a manual test run with this plus https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/156 and https://github.com/charmed-kubernetes/layer-kubernetes-common/pull/17 cherry-picked in, and a worker built from https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/84 pre-deployed into the model.  Note that to get the test to work with the pre-deployed worker, you have to use the master branch of python-libjuju until the next version is released to PyPI:

```
tox -re integration --notest
.tox/integration/bin/pip uninstall juju
.tox/integration/bin/pip install --no-cache https://github.com/juju/python-libjuju/archive/refs/heads/master.zip#egg=juju
```

johnsca:
New [manual integration test results](https://pastebin.ubuntu.com/p/2DyVVhyTNd/) after rebase from master.

Cynerva:
I walked through the upgrade path for this. The port served by kubeapi-load-balancer changed from 443 to 6443, which meant I had to redownload the kubeconfig from kubernetes-master. Make sure to include that in the documentation upgrade notes.

reactive/kubernetes_master.py
```diff
@@ -1405,25 +1409,62 @@ def create_tokens_and_sign_auth_requests():
...
+def request_load_balancers():
+    """Request LBs from the related provider(s)."""
+    for lb_type in ("internal", "external"):
+        lb_provider = endpoint_from_name("loadbalancer-" + lb_type)
+        if not lb_provider.is_available:
+            continue
+        req = lb_provider.get_request("api-server-" + lb_type)
+        req.protocol = req.protocols.tcp
+        api_port = kubernetes_master.STANDARD_API_PORT
+        req.port_mapping = {api_port: api_port}
```
Cynerva, line 1440: It might be worth considering mapping 443 to api_port since that's how it has been with kubeapi-load-balancer in the past.

Cynerva (approved):
LGTM

PR #154, opened 2021-04-15
Bump release version in default config and tests

kwmonroe:
1.21 is now stable. Edge builds from master branch should move up to 1.22.

Test bundle should use use 1.21/stable.  I got the resource IDs for the test updates with the following:
```bash
...
$ charm show ~containers/kubernetes-worker id resources -c stable
id:
  Id: cs:~containers/kubernetes-worker-743
  Name: kubernetes-worker
  Revision: 743
  User: containers
resources:
- Description: CNI plugins for amd64
  Fingerprint: kuelBRY0tv1sH8l8FVIg5g6kcG8yimyr9mLc/96vPag+GeriIw9zj8dZtKNYNlbr
  Name: cni-amd64
  Path: cni.tgz
  Revision: 747
...
```

### DIFF
--- a/tests/data/bundle.yaml
+++ b/tests/data/bundle.yaml
@@ -10,12 +10,9 @@ machines:
 services:
   containerd:
     charm: cs:~containers/containerd
-    resources: {}
   easyrsa:
     charm: cs:~containers/easyrsa
     num_units: 1
-    resources:
-      easyrsa: 5
     to:
     - '1'
   etcd:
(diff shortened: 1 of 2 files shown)


Cynerva (changes requested):
> Test bundle should use use 1.21/stable.

Why 1.21/stable here, but 1.22/edge in the default config? If it's too early to use 1.22/edge in the tests, then it's too early to set it in default config also.

> I got the resource IDs for the test updates with the following:

I don't think resource revs make sense in that bundle. We pin the resource revs, but not the charm revs? I would just remove the resource revs from the bundle entirely. Then it will use the latest resources from whichever channel the charms are deployed from. I'm pretty sure that's what we want.

tests/data/bundle.yaml
```diff
@@ -10,46 +10,27 @@ machines:
...
     - '0'
   flannel:
     charm: cs:~containers/flannel
-    resources:
-      flannel-amd64: 653
-      flannel-arm64: 650
-      flannel-s390x: 637
   kubernetes-master:
-    charm: {{master_charm}}
+    charm: cs:~containers/kubernetes-master
```
Cynerva, line 28: What happened to `{{master_charm}}`? I think the integration tests need that to deploy a locally built master instead of pulling it from the charm store: https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/91e7ae2543db2629504ab6efd9ecac8687396fa0/tests/integration/test_kubernetes_master_integration.py#L11-L13

Cynerva (changes requested):

tests/data/bundle.yaml
kwmonroe: gah!  they absolutely do.  i modified that for a local deployment to make sure i could just drop all resources from the test bundle.yaml.  forgot to put it back :/

Cynerva (approved):
LGTM, thanks!

PR #155, opened 2021-04-21
Add experimental docs

evilnick:


evilnick:
This PR adds a docs directory and an experimental single page docs version of the k8s-master docs. This is mainly so I can proceed with developing the scripts to sync back and forth with discourse, and work out which bits will need to be rewritten.

chris-sanders:
@evilnick this looks good as a start. I'm a bit horrified by the mix of HTML in markdown but it's a start :)

chris-sanders (approved):

evilnick:
The HTML is the only way, currently, we can make the tables work for the Config options. The templates there will need to be rewritten anyhow as the docs pages render in a different way from discourse, sadly. So, don't be too worried about the content anyhow.
Oh, and all the HTML is autogenerated by the scripts - no need for people to edit/write manually or even touch those sections ;)

PR #157, opened 2021-05-04
Remove dropped kube-apiserver arg

johnsca:
The `--kubelet-https` arg has been [deprecated][] since 1.19 and has been dropped entirely in 1.22, breaking the deployment.

[deprecated]: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#other-cleanup-or-flake-5


johnsca:
Since the GH runners are disabled for the time being, [here is a passing integration run](https://pastebin.canonical.com/p/q4mS8YM4tQ/).

johnsca:
I have no idea why the lint checks are failing.  It has nothing to do with the PR and seems like it might be an issue with the checkout action, since it's [getting the wrong commit SHA](https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/157/checks?check_run_id=2504624249#step:2:232).

kwmonroe (approved):

PR #158, opened 2021-05-12
Fix Focal deploys broken by Werkzeug 2.0

johnsca:
Yesterday (May 11), Werkzeug 2.0.0 was released which includes a conditional dependency of dataclasses for Python <3.7 but that fails to install on 3.7+. Because the charm snap uses Python 3.6 during the build but Focal includes Python 3.8, deployments there are now broken.

This pins the Werkzeug library and adds tests to validate that the built wheelhouse will successfully install on a given Python version.

Fixes: [lp:1928254][]

[lp:1928254]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1928254


.github/workflows/main.yml
```diff
@@ -18,26 +18,13 @@ jobs:
...
       run: tox -vve lint
-  unit-test:
-    name: Unit Tests
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python: [3.6, 3.7, 3.8, 3.9]
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v2
```
chris-sanders, line 31: Is this not necessary to get the repository checked out for the tox runs?

chris-sanders (commented):
One question, if that's fine then I'm good.

.github/workflows/main.yml
johnsca: I consolidated the lint & unit (and now the wheelhouse) tasks into a single one to speed up the checks on GH (not blocking on separate VMs) and reduce our resource usage.  It's hidden by the upper fold, but that part is still there in the workflow.

chris-sanders: Oh wait this is just the diff. There is a checkout it's earlier in the code. You combined lint, unit, wheelhouse so the diff just shows the removal b/c the checkout already existed on the lint step.

chris-sanders (approved):
Looks good

johnsca:
Backported to stable in 0daa2c3

PR #160, opened 2021-05-18
Use ensure_future instead of create_task

johnsca:
The `create_task()` function was only added to `asyncio` in 3.7, so bionic and before need to use the older (but equivalent) `ensure_future()`.

Fixes [lp:1928737](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1928737)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,8 +36,12 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


kwmonroe (approved):
This looks good.  @johnsca and I chatted about the additional vsphere params that aren't related to the py < 37 fix, but it feels like a fine drive-by that we need for gh actions in this repo.

johnsca:
This was backported to stable in ae9bf6f

PR #161, opened 2021-05-19
Improve status reporting during deploy

johnsca:
Certain issues during deploy, such as the API server failing to start, can be misreported in status as "Failed to to setup auth-webhook tokens" rather than a more specific and useful message.

Fixes [lp:1928951][]

[lp:1928951]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1928951


johnsca:
Ideally, some of these (like the API server) would be fatal, but occasionally things can be slow enough that it's worth waiting for it to retry on the next hook. Maybe worth reconsidering some of them, though.

Cynerva (approved):

PR #162, opened 2021-06-10
Added support for new enable-csi-snapshots option in cdk-addons

Vultaire:
This is an attempt (perhaps naive) at addressing https://bugs.launchpad.net/cdk-addons/+bug/1926494.

Dependencies:
* Patch to cdk-addons: https://github.com/charmed-kubernetes/cdk-addons/pull/206
* rocks.canonical.com:443/cdk does not have a copy of k8s.gcr.io/sig-storage/snapshot-controller:v3.0.2, and so the volume-snapshot-controler StatefulSet's pods will not spin up correctly right now unless the image is patched to look at the k8s.gcr.io version.

Also, I acknowledge that there are no tests for this.  I've reviewed the unit and integration tests currently present for this charm, and while technically I could write an integration test for this, it wouldn't really match the scope of the tests which presently exist here, so it didn't quite feel appropriate.  However, if you have suggestions for a test case you think would be suitable, I'm happy to do it; please let me know.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -588,6 +588,18 @@ def enable_metric_changed():
         configure_cdk_addons()
 
 
+@when("etcd.available")
+@when("config.changed.enable-csi-snapshots")
+def enable_csi_snapshots_changed():
+    """
+    Trigger CDK addon configuration if enable-csi-snapshots is updated.
+
+    :return: None
+    """
+    if is_state("leadership.is_leader"):
+        configure_cdk_addons()
+
+
 @when("config.changed.client_password", "leadership.is_leader")
 def password_changed():
     """Handle password change by reconfiguring authentication."""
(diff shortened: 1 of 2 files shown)


reactive/kubernetes_master.py
```diff
@@ -588,6 +588,18 @@ def enable_metric_changed():
...
+@when("etcd.available")
+@when("config.changed.enable-csi-snapshots")
+def enable_csi_snapshots_changed():
+    """
+    Trigger CDK addon configuration if enable-csi-snapshots is updated.
+
+    :return: None
+    """
+    if is_state("leadership.is_leader"):
+        configure_cdk_addons()
```
Cynerva, line 600: configure_cdk_addons runs on every hook anyway. I do not think this enable_csi_snapshots_changed handler would be needed at all.

Cynerva (commented):
See discussion in https://github.com/charmed-kubernetes/cdk-addons/pull/206 about long-term viability of adding this to cdk-addons. I do not think we will end up accepting this PR.

PR #163, opened 2021-06-14
Remove auth secret functions, don't set kubeconfig_path

Cynerva:
https://bugs.launchpad.net/charm-calico/+bug/1920034

The auth secret functions have been moved to layer-kubernetes-common (see https://github.com/charmed-kubernetes/layer-kubernetes-common/pull/19), so remove them here.

Also notify CNI subordinates when the admin kubeconfig has changed. Don't send kubeconfig_path which is no longer needed (see https://github.com/juju-solutions/interface-kubernetes-cni/pull/9).

I don't think any test changes are needed for this one, but it can't be merged until other PRs are ready.


Cynerva:
This should be merged with https://github.com/juju-solutions/interface-kubernetes-cni/pull/9.

Test failure is because https://github.com/juju-solutions/interface-kubernetes-cni/pull/9 isn't merged yet, but these really have to be merged at the same time or we're gonna have broken edge charms.

kwmonroe (approved):

PR #164, opened 2021-06-23
Fix NoneType error when using old loadbalancer relation

johnsca:
If the legacy `loadbalancer` relation is in `goal-state` but not yet ready, the following error can occur:

```
  File "lib/charms/layer/kubernetes_master.py", line 139, in get_external_api_endpoints
    lb_addresses = loadbalancer.get_addresses_ports()
AttributeError: 'NoneType' object has no attribute 'get_addresses_ports'
```

This fixes it by making that section use `endpoint_from_name` like all the other calls in that function and its internal counterpart.

### DIFF
--- a/lib/charms/layer/kubernetes_master.py
+++ b/lib/charms/layer/kubernetes_master.py
@@ -135,7 +135,7 @@ def get_external_api_endpoints():
 
     # Support the older loadbalancer relation (public-address interface).
     if "loadbalancer" in goal_state["relations"]:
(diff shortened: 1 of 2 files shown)


johnsca:
Originally encountered [in this test failure](https://github.com/charmed-kubernetes/charm-kubeapi-load-balancer/pull/12/checks?check_run_id=2898096752#step:5:396).

kwmonroe (approved):

PR #166, opened 2021-07-12
Remove config/actions deprecated prior to 1.19

kwmonroe:
In an effort to keep our code base tidy, remove config and actions that were deprecated prior to 1.19.

~~Draft until LP is open.~~

Fixes [LP:1935992](https://bugs.launchpad.net/charm-kubernetes-worker/+bug/1935992) by removing:

- addons-registry config
- create-rbd-pv action and related templates
- monitoring-storage config
- kube-dns interface
- migrate_from_pre_snaps code


actions.yaml
```diff
@@ -1,31 +1,5 @@
 restart:
   description: Restart the Kubernetes master services on demand.
-create-rbd-pv:
```
Cynerva, line 3:
```
$ git grep create-rbd-pv
docs/index.md:      create-rbd-pv
templates/rbd-persistent-volume.yaml:# `create-rbd-pv` action. This is a temporary file on disk to enlist resources.
```

Looks like there's a doc reference to create-rbd-pv that needs to be removed.

It also looks like `templates/rbd-persistent-volume.yaml` was only used by the create-rbd-pv action and can be removed. Yay! Mind doing that?

Cynerva (changes requested):

Cynerva (approved):

PR #169, opened 2021-07-29
Allow custom layers / interfaces to be used in integration tests

freyes:
`charm build` will pull layers and interfaces from the index by default,
but when a patch depends on changes non upstreamed to test locally the
environment variables CHARM_INTERFACES_DIR and CHARM_LAYERS_DIR need to
be set to the path where the interfaces and layers are in the local
machine.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,8 @@ envlist = lint,unit,integration
 setenv =
     PYTHONPATH={toxinidir}:{toxinidir}/lib
     PYTHONBREAKPOINT=ipdb.set_trace
+    CHARM_LAYERS_DIR={toxinidir}/layers
+    CHARM_INTERFACES_DIR={toxinidir}/interfaces
 
 [testenv:unit]
 deps =


freyes:
@Cynerva hi, this is the change we discussed yesterday. Thanks.

Cynerva:
I think it was @johnsca who discussed this with you actually. It looks good to me. @johnsca you wanna confirm this is what you had in mind?

johnsca:
My concern with permanently adding this to the repo is that it will mean any tests run locally will potentially use whatever local layers one might have checked out, on whatever random branch. Maybe this isn't too big of a concern, since I don't typically think about it when doing local manual tests, either, but I wonder if it would at least be worth adding some logic to pytest-operator to exclude those by default with an arg to override that?

freyes:
This is what the openstack team does (for charm-octavia) https://opendev.org/openstack/charm-octavia/src/branch/master/tox.ini#L19 , so if you want/need to override something you create a directory inside repo so you shouldn't see pollution from the machine where the tests are run

johnsca:
I really like that.  Would you mind changing the PR to use that pattern instead?

freyes:
@johnsca done, updated the patch, I believe this way is better too specially since we would be keeping some degree of consistency with the OpenStack charms.

johnsca (approved):
Great.  This provides a way to easily use modified layers without accidentally doing so.

PR #170, opened 2021-07-29
Move secrets refresh task out of auth-webhook app startup

johnsca:
We see somewhat frequent issues with gunicorn killing auth-webhook workers after claiming that they failed to start, with no apparent reason as to why they might be considered to have failed to start. This attempts to address this by moving the secret refresh task entirely out of the startup process and instead kick it off on the first request. This does mean that that first request will need to wait a bit for the secrets to be loaded, so some logic was added to handle the case where the secrets are not available but are actively being fetched.


johnsca:
Closed in favor of #171 which is a less invasive fix.

PR #171, opened 2021-07-29
Fix auth-webhook crash due to gunicorn SIGCHLD madness

Cynerva:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1938470. More details there.


johnsca (approved):
LGTM

Cynerva:
Verified locally that with this fix, auth-webhook no longer crashes while kube-apiserver is down.

PR #172, opened 2021-08-17
CK-124 - Fix port for loadbalancer-external

johnsca:
The old external LB support using kubeapi-load-balancer and the old public-address interface used the default config value for the port from the ka-l-b charm of 443. The switch to the new loadbalancer interface ended up unintentionally switching that in the default configuration to use the standard API port of 6443 which broke the test_sans test case.


johnsca:
Local passing test run (with the additional info from https://github.com/charmed-kubernetes/jenkins/pull/724):

```
jobs/integration/validation.py::test_sans 
-------------------------------- live log setup --------------------------------
unknown common facade version for CAASUnitProvisioner
  Checking for example domain added to certs...
  Checking for example domain removed from certs...
  Example domain still in cert for kubernetes-master/0
  Example domain still in cert for kubernetes-master/1
  Checking for example domain removed from certs...
PASSED
```

kwmonroe (approved):

PR #180, opened 2021-09-17
Fix action branch and wheelhouse validation

johnsca:
Branch for actions-operator was renamed, and setuptools released a breaking change that only affects the validate-wheelhouse test, not the actual charm deploys (since the base VMs for deploys won't have the newer setuptools).


kwmonroe (approved):
Arguably my most "yuk" +1.  Branch rename is legit, pin for setuptools is `ugh`.

PR #182, opened 2021-10-27
LP#1936816 - auth-webhook.pid to /run/ and auth-webhook.log to /var/log

addyess:
* address bug: move auth-webhook.pid to /run rather than /root/cdk
* address bug: move auth-webhook.log to `/var/log/kubernetes`
* move existing auth-webhook logs to `/var/log/kubernetes`
* update logrotated configuration to appropriate rotate the logs in their new home
* add NRPE check to ensure the `cdk.master.auth-webhook` service is running

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -130,7 +130,9 @@ aws_iam_webhook = "/root/cdk/aws-iam-webhook.yaml"
 auth_webhook_root = "/root/cdk/auth-webhook"
 auth_webhook_conf = os.path.join(auth_webhook_root, "auth-webhook-conf.yaml")
 auth_webhook_exe = os.path.join(auth_webhook_root, "auth-webhook.py")
-auth_webhook_svc = "/etc/systemd/system/cdk.master.auth-webhook.service"
+auth_webhook_svc_name = "cdk.master.auth-webhook"
+auth_webhook_svc = "/etc/systemd/system/{}.service".format(auth_webhook_svc_name)
+
 
 register_trigger(
     when="endpoint.aws.ready", set_flag="kubernetes-master.aws.changed"  # when set
@@ -1017,9 +1019,9 @@ def add_systemd_file_watcher():
 @when("etcd.available", "tls_client.certs.saved")
 @restart_on_change(
     {
-        auth_webhook_conf: ["cdk.master.auth-webhook"],
-        auth_webhook_exe: ["cdk.master.auth-webhook"],
-        auth_webhook_svc: ["cdk.master.auth-webhook"],
+        auth_webhook_conf: [auth_webhook_svc_name],
+        auth_webhook_exe: [auth_webhook_svc_name],
+        auth_webhook_svc: [auth_webhook_svc_name],
     }
 )
 def register_auth_webhook():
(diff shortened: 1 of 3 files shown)


reactive/kubernetes_master.py
```diff
@@ -126,11 +127,14 @@
 kubecontrollermanagerconfig_path = "/root/cdk/kubecontrollermanagerconfig"
 kubeschedulerconfig_path = "/root/cdk/kubeschedulerconfig"
 cdk_addons_kubectl_config_path = "/root/cdk/cdk_addons_kubectl_config"
+cdk_logs = "/var/log/cdk/"
```
kwmonroe, line 130: Similar comment as https://github.com/charmed-kubernetes/kubernetes-docs/pull/617#pullrequestreview-795425141:

I wonder since we're making this change if we should drop cdk in favor of ck. ck is awfully short and could confuse people wondering what's this ck subdirectory for? so perhaps we should consider charmed-kubernetes as our subdir name and ck in filenames?

reactive/kubernetes_master.py
```diff
@@ -1076,12 +1081,19 @@ def register_auth_webhook():
...
     render("cdk.master.auth-webhook-conf.yaml", auth_webhook_conf, context)
     render("cdk.master.auth-webhook.py", auth_webhook_exe, context)
     render(
         "cdk.master.auth-webhook.logrotate", "/etc/logrotate.d/auth-webhook", context
     )
 
+    # Move existing log files from ${auth_webhook_root} to /var/log/cdk/
+    for log_file in Path(auth_webhook_root).glob("auth-webhook.log*"):
+        if not (k8s_log_path / log_file.name).exists():
+            shutil.move(str(log_file), str(k8s_log_path))
```
Cynerva, line 1095: The new log file is `/var/log/kubernetes/cdk.master.auth-webhook.log`, but the move here places the old log file at `/var/log/kubernetes/auth-webhook.log`. Those are different - is this intentional?

Cynerva (changes requested):
Need clarification, see comment.

reactive/kubernetes_master.py
addyess: It's a very good question.  
I did this intentionally in the event that `filebeat` or other log collecting services would "see" the old file and collect its logs if it wasn't already collected.  I feared restarting the service while the file was in one location or the other.  this was an attempt to just "carry over" any logs or rotated during a charm upgrade without being destructive.

I'd value your opinion here though

Cynerva: I guess my concern is that when someone comes along and looks in /var/log/kubernetes, on an upgraded cluster they might see:

```
$ ls -l /var/log/kubernetes
total 3784
-rw-r--r-- 1 root root   34138 Nov  5 20:17 auth-webhook.log
-rw-r--r-- 1 root root 3515014 Nov  8 16:37 cdk.master.auth-webhook.log
```

They might be confused about the difference between these two files, until they look at the dates and realize that auth-webhook.log is no longer being written. They have to *infer* that auth-webhook.log is an old file that's no longer being touched.

I would be tempted to add the `cdk.master.` prefix when moving the logs, to make the transition more seamless.

addyess: A rename sounds in order.  

I'll test to make sure that the move|rename happens in such a way we do no overwrite a log currently in use by a an active service writing to it.

addyess: Before Upgrade:
```bash
root@juju-958d60-0:~/cdk/auth-webhook# ls
__pycache__             auth-webhook.log       auth-webhook.log.10.gz  auth-webhook.log.3.gz  auth-webhook.log.5.gz  auth-webhook.log.7.gz  auth-webhook.log.9.gz
auth-webhook-conf.yaml  auth-webhook.log.1.gz  auth-webhook.log.2.gz   auth-webhook.log.4.gz  auth-webhook.log.6.gz  auth-webhook.log.8.gz  auth-webhook.py
```

After Upgrade:
```bash
root@juju-958d60-0:~/cdk/auth-webhook# ls
__pycache__  auth-webhook-conf.yaml  auth-webhook.py
root@juju-958d60-0:~/cdk/auth-webhook# ls /var/log/kubernetes/
cdk.master.auth-webhook.log       cdk.master.auth-webhook.log.10.gz  cdk.master.auth-webhook.log.3.gz  cdk.master.auth-webhook.log.5.gz  cdk.master.auth-webhook.log.7.gz  cdk.master.auth-webhook.log.9.gz
cdk.master.auth-webhook.log.1.gz  cdk.master.auth-webhook.log.2.gz   cdk.master.auth-webhook.log.4.gz  cdk.master.auth-webhook.log.6.gz  cdk.master.auth-webhook.log.8.gz
```

Cynerva (approved):

PR #184, opened 2021-11-22
Remove deprecated call to CNI relation set_config

Cynerva:
CNI subordinate behavior will no longer differ between masters and workers. As such, `is_master` is being removed from the CNI relation, and since there is no other config to set, `cni.set_config` is going away too. So this just cleans up nicely.

This is needed to support adding kubelet to kubernetes-master.

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -548,14 +548,6 @@ def storage_backend_changed():
     remove_state("kubernetes-master.components.started")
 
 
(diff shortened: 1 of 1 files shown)


addyess:
LGTM :+1:

addyess:
you probably need to rebase with master

kwmonroe (approved):
LGTM; failing test is fixed in #188.

PR #189, opened 2021-12-06
Add CNI resources to integration tests

addyess:


tests/integration/test_kubernetes_master_integration.py
```diff
@@ -21,15 +22,51 @@ def _check_status_messages(ops_test):
...
+    await _retrieve_url(CNI_AMD64_URL, cni_amd64)
+
+    log.info("Build Charm...")
+    charm = await ops_test.build_charm(".")
+
+    # Work around libjuju not handling local file resources by manually
+    # pre-deploying the charm w/ resource via the CLI. See
+    # https://github.com/juju/python-libjuju/issues/223
+    log.info("Deploy Charm...")
+    rc, stdout, stderr = await ops_test._run(
```
Cynerva, line 47: I know this is basically a copy/paste of what kubernetes-worker integration tests do. But it seems like it would be simpler to just `juju deploy` the entire bundle via CLI, so we don't have to do this whole pre-deploy thing.

tests/integration/test_kubernetes_master_integration.py
```diff
@@ -21,15 +22,51 @@ def _check_status_messages(ops_test):
...
+            with target_file.open("wb") as fd:
+                async for chunk in resp.content.iter_chunked(chunk_size):
+                    fd.write(chunk)
+
+
 @pytest.mark.abort_on_fail
 async def test_build_and_deploy(ops_test):
+    log.info("Retrieve CNI...")
+    cni_amd64 = ops_test.tmp_path / "cni-amd64.tgz"
+    await _retrieve_url(CNI_AMD64_URL, cni_amd64)
```
Cynerva, line 38: Long-term, I would really like to see the integration tests build the CNI resources rather than pulling a pre-built one.

Cynerva (approved):
I see room for improvement in this code but I'm good with it as-is for getting the tests unblocked. Thanks for doing this.

tests/integration/test_kubernetes_master_integration.py
addyess: totally agree, i could make that change pretty quickly

tests/integration/test_kubernetes_master_integration.py
addyess: i have no trouble with this.

Cynerva (approved):

PR #190, opened 2021-12-06
HACluster - Removing master services from Pacemaker and leaving to Systemd

gabrielcocenza:
Currently, charm-hacluster is not fully prepared to receive kubernetes master services
for some reasons:

1. charm-hacluster doesn’t wait to receive all constraints of order and colocation necessary to
run master services without conflict. Kube-apiserver should be running in a node before
Pacemaker tries to start kube-controller-manager in the same node. [1]

2. Master services are cloned (running in all nodes) with Pacemaker and not grouped with VIP
resources, meaning that if a resource starts to fail in a node that hosts the VIP, it won’t try to
change node leaving the VIP in a unit where it’s not healthy. [2]

3. Resources added with the hacluster-interface have the `migration-threshold` set to
`INFINITY`. This means that the cluster will try to recover the service in the node forever. In a
scenario where just restarting the service is not able to recover the service, Pacemaker won’t
change the VIP to another node even if it’s grouped with the master services. [3]

Knowing those issues, the short fix proposed in this PR is to remove the master services from
Pacemaker when hacluster is present in the bundle and leave to systemd that will try to restart if
the service fails. The `atexit` checks the health of master services and if some service is not
running, it will set the node in `standby` mode and this will force migration of VIP to another
node.

To make those changes take effect, first it’s necessary to approve the layer-hacluster PR [4].

One known problem of this approach is that sometimes kubernetes-master hooks get "stuck" in
a busy loop and never exit which would affect the logic needed in `atexit`. Besides this, it could
take up to 5 minutes to identify the problem and change the VIP to another node in the cluster.
The long-term solution needs to solve problems 1, 2 and 3.  

Closes-bug: [#1929234](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1929234)

[1] https://bugs.launchpad.net/charm-hacluster/+bug/1952492
[2] https://bugs.launchpad.net/charm-interface-hacluster/+bug/1952753
[3] https://github.com/openstack/charm-interface-hacluster/blob/master/interface_hacluster/common.py#L993-L1009
[4] https://github.com/juju-solutions/layer-hacluster/pull/3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ __pycache__/
 *.pyc
 placeholders/
 *.charm
+interfaces
+layers
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -25,7 +25,7 @@ import yaml
 from itertools import filterfalse
 from shutil import move, copyfile
 from pathlib import Path
-from subprocess import check_call
+from subprocess import check_call, call
 from subprocess import check_output
 from subprocess import CalledProcessError
 from urllib.request import Request, urlopen
@@ -57,8 +57,6 @@ from charmhelpers.contrib.charmsupport import nrpe
 from charms.layer import kubernetes_master
 from charms.layer import kubernetes_common
 
-from charms.layer.hacluster import add_service_to_hacluster
-from charms.layer.hacluster import remove_service_from_hacluster
 from charms.layer.kubernetes_common import kubeclientconfig_path
 from charms.layer.kubernetes_common import migrate_resource_checksums
 from charms.layer.kubernetes_common import check_resources_for_upgrade_needed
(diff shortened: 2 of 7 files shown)


Cynerva (changes requested):
I have two concerns with this PR:

1) Upgraded deployments will be coming from an earlier version of the charm that already added services to hacluster. I don't see any code to remove them after upgrading. I think there needs to be code in the upgrade-charm hook handler to remove the services from hacluster.

2) I don't think it makes sense to have the integration tests always deploy hacluster. We want to be able to run the integration tests in a variety of environments, and in particular the GitHub actions run them in a vSphere environment that doesn't have addresses to spare for VIPs. The hacluster part needs to be made optional, along with the test_service_down that goes with it.

Cynerva (changes requested):
Thanks for addressing my earlier concerns.

Just one last request: `tests/data/overlays/hacluster.yaml` is not really an overlay. Can you please move or rename that file to make it clear that it's a Bundle, not an Overlay? Something like `tests/data/bundle-hacluster.yaml` or `tests/data/bundles/hacluster.yaml` should be fine.

gabrielcocenza:
Thanks for your review.

I agree that the file is a bundle instead of an `overlay`. I've changed to `tests/data/bundle-hacluster.yaml` as you pointed.

Cynerva (approved):
Thanks. LGTM

PR #193, opened 2021-12-14
Disable CSIMigration* for control plane services

Cynerva:
In testing 1.23, SQA encountered issues creating PVCs on AWS. This was caused by the CSIMigrationAWS feature that was set to true in the k8s 1.23 release. The feature depends on AWS EBS CSI drivers that we do not ship.

Disable it for now, along with CSIMigrationGCE and CSIMigrationAzureDisk since we are likely to hit the same problem with those features too.


Cynerva:
Fixes https://bugs.launchpad.net/bugs/1954807

addyess (approved):
LGTM :+1:

PR #202, opened 2022-02-15
resolve race conditions on the external-cloud-provider relation being available

addyess:


johnsca (approved):
Much nice, very better.

PR #197, opened 2022-01-13
inclusive naming for kubernetes-master

addyess:
the charm kubernetes-master is being renamed to kubernetes-control-plane and is available on [charmhub.io](https://charmhub.io/kubernetes-control-plane)

The following PR updates the charm name as well as uses associated with the integration testing.  It doesn't address all inclusive naming issues within the charm itself -- but just attempts to alter the name of the charm for maximizing backwards compatability

PR #187, opened 2021-12-02
Existing tests pass with the inclusion of keystone

addyess:
Adding keystone and related components to integration tests bundle:
`keystone`
`keystone-mysql-router`
`mysql-innodb-cluster`

updated tests to use the `juju-context` from the generated kubeconfig

PR #203, opened 2022-02-24
Stable cherries

kwmonroe:
We need to bring the bits related to renaming k8s-master -> k8s-control-plane into `stable`.  Otherwise, we can't build a k8s-control-plane charm from the `stable` branch.


addyess:
how'd you decide which cherries to pick?

kwmonroe:
I chose the #194 cherry because that had a 1.23-ck1 target.

I chose the #187 cherry because that altered k-c-p unit tests significantly enough to make inclusion easier than merging.

I chose #197 because that's the actual rename that we need to make this bugfix release possible from charmhub (ch *only* has k-c-p, there is no kubernetes-master charm there).

Cynerva (approved):
Oh, thank you. I was working on this today and hadn't seen this PR. I actually chose the same exact commits you did, so, I'm glad we're on the same page here. :+1: LGTM

PR #204, opened 2022-03-03
configure node-labels on control plane nodes

addyess:
Use with the following PRs:
* https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/110
* https://github.com/charmed-kubernetes/layer-kubernetes-master-worker-base/pull/16


reactive/kubernetes_master.py
```diff
@@ -3658,11 +3696,73 @@ def configure_kubelet():
     kubernetes_common.configure_kubelet(
         dns_domain, dns_ip, registry, taints=taints, has_xcp=has_xcp
     )
+    set_state("kubernetes-master.config.created")
     service_restart("snap.kubelet.daemon")
-
+    set_state("kubernetes-master.label-config-required")
```
Cynerva, line 3701: Do we need to reconfigure labels every time Kubelet restarts? Once the node has been registered, Kubelet shouldn't be changing its own labels.

I think really, we just need to set the labels one time after the Node has been created. Then update them any time the charm is upgraded (in case the charm's label handling code changed) or when the labels config changed.

reactive/kubernetes_master.py
```diff
@@ -3658,11 +3696,73 @@ def configure_kubelet():
...
     set_flag("kubernetes-master.kubelet.configured")
 
 
+@when("config.changed.labels")
+def handle_labels_changed():
+    set_state("kubernetes-master.label-config-required")
+
+
+@when("kubernetes-master.label-config-required", "kubernetes-master.config.created")
+def apply_node_labels():
```
Cynerva, line 3711: I would wait for `kubernetes-master.apiserver.configured` here as well. We don't want to try to talk to an apiserver that's not up yet.

Cynerva: Also `kubernetes-master.kubelet.configured`. We don't want to try to label a node that doesn't exist yet.

Cynerva: If you make the `kubelet_kubeconfig_path` -> `kubeclientconfig_path` change I suggested in prior comments, I suspect you could also replace `kubernetes-master.config.created` here with `authentication.setup` since that's what signals the creation of the client kubeconfig [here](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/1853eb192065c27453b234c0b50ebf6093f4d790/reactive/kubernetes_master.py#L2231-L2237).

Cynerva (changes requested):

reactive/kubernetes_master.py
```diff
@@ -3658,11 +3696,73 @@ def configure_kubelet():
     kubernetes_common.configure_kubelet(
         dns_domain, dns_ip, registry, taints=taints, has_xcp=has_xcp
     )
+    set_state("kubernetes-master.config.created")
```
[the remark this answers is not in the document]
addyess: I really don't know.  I stole this from [k8s-worker](https://github.com/charmed-kubernetes/charm-kubernetes-worker/blob/master/reactive/kubernetes_worker.py#L628).  Do you have another suggestion?

reactive/kubernetes_master.py
addyess: maybe?  again... was just following what k8s-worker was doing

reactive/kubernetes_master.py
addyess: Thanks for this review.  @kwmonroe was suggesting i push these changes down into kubernetes-master-worker-base such that they share much of this work.  Maybe I could still use similar reactive states to trigger the node label application

Cynerva (approved):

PR #205, opened 2022-03-04
Service cidr

redno2:
By default, a service with a /24 is clearly not enough for a k8s cluster, it should be changed to a /16

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -80,7 +80,7 @@ options:
       created for the master nodes.
   service-cidr:
     type: string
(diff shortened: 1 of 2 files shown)


kwmonroe:
@redno2 thank you for the contribution!  We're closing in on the 1.24 release and have decided this will be too risky to include.  I've opened the following to track this for inclusion in our next release:

https://bugs.launchpad.net/charm-kubernetes-master/+bug/1968071

PR #207, opened 2022-03-21
handle when kubelet service fails to heal

addyess:
Addresses [LP1965765](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1965765) by giving the charm some action to take in the event the kubelet service fails to start on the control-plane node

### DIFF
--- a/reactive/kubernetes_master.py
+++ b/reactive/kubernetes_master.py
@@ -3725,4 +3725,5 @@ HEAL_HANDLER = {
         "run": start_master,
         "clear_flags": ["kubernetes-master.components.started"],
     },
+    "kubelet": {"run": reconfigure_kubelet, "clear_flags": []},
 }


reactive/kubernetes_master.py
```diff
@@ -3725,4 +3725,5 @@ def configure_default_cni():
         "run": start_master,
         "clear_flags": ["kubernetes-master.components.started"],
     },
+    "kubelet": {"run": reconfigure_kubelet, "clear_flags": []},
```
addyess, line 3728: @Cynerva is this the right function to run to reconfigure the kubelet

https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/207/files#diff-b3099d2217027f2f460740149252f0dca742dbb5deeb274bc17559fb648b1d86R3685-R3692

addyess: it seemed appropriate but I wasn't totally sure

Cynerva: :+1: Looks reasonable to me

PR #208, opened 2022-03-28
create inclusive naming check

addyess:
So, there ARE inclusive naming failures but they've been filtered by reviewdog because essentially we're not introducing new failures -- only eliminating existing ones.  I still need to manually test an upgrade from a previous deployment.  Diligent reviews are appreciated

Merge first
* https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/210
* https://github.com/charmed-kubernetes/layer-kubernetes-node-base/pull/1
* https://github.com/juju/layer-index/pull/123

Merge along side:
* https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/114

PR #211, opened 2022-04-05
Update kubernetes_control_plane.py

addyess:
Fix missed suggestions about renaming flags from @kwmonroe


kwmonroe (approved):

kwmonroe:
Follow-on to #208

PR #215, opened 2022-04-18
always enable apiserver flags, regardless of enable-metrics settings

addyess:


kwmonroe (approved):
LGTM; please link an LP bug for k-c-p 1.24 so this gets bubbled up into ck 1.24 release notes.

PR #219, opened 2022-04-21
Use ceph-client relation instead of ceph-admin

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1840260 and https://bugs.launchpad.net/bugs/1840492 for kubernetes-control-plane.
Depends on https://github.com/charmed-kubernetes/cdk-addons/pull/212.

The ceph-storage (ceph-admin) relation that we currently use is considered experimental, and is not available when relating to the ceph-proxy charm.

This PR drops all usage of the ceph-storage relation, in favor of using ceph-client instead, which is mature and widely available.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -1000,6 +1000,12 @@ def set_final_status():
         hookenv.status_set("active", msg)
         return
 
+    if is_flag_set("ceph-storage.available"):
+        hookenv.status_set(
+            "blocked", "ceph-storage relation deprecated, use ceph-client instead"
+        )
+        return
+
     if is_flag_set("ceph-client.connected") and not is_flag_set(
         "ceph-client.available"
     ):
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -53,6 +53,7 @@ from charmhelpers.core.host import restart_on_change
 from charmhelpers.core.host import service_pause, service_resume, service_stop
 from charmhelpers.core.templating import render
 from charmhelpers.contrib.charmsupport import nrpe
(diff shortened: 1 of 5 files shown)


reactive/kubernetes_control_plane.py
```diff
@@ -1809,60 +1808,7 @@ def addons_ready():
         return False
 
 
-@when("ceph-storage.available")
-def ceph_state_control():
```
addyess, line 1813: Could you also `block` on this hook to inform folks not to use this relation?  In the case of an upgrade to this charm -- i think i'd wanna know why this relation isn't doing what it used to

addyess: also, are there any flags that can be removed/cleared if the `ceph-storage` relation is disconnected

addyess (approved):
Looking at the logic, all looks well.  I'll read through it again and see If i can spot anything

reactive/kubernetes_control_plane.py
Cynerva: > Could you also `block` on this hook to inform folks not to use this relation? In the case of an upgrade to this charm -- i think i'd wanna know why this relation isn't doing what it used to

Good idea. I've added a blocked status message for it.

> also, are there any flags that can be removed/cleared if the ceph-storage relation is disconnected

I don't think so. It looks like the only relevant flag is `ceph-storage.available` but the ceph-admin interface clears it on departure [here](https://github.com/openstack-charmers/juju-interface-ceph-admin/blob/c1061a29297084fa53c2474ba371671186ff3389/requires.py#L6-L11).

kwmonroe (approved):
+1, big fan of seeing those conditions around `1.12` go away too!

PR #220, opened 2022-04-25
Add calls to write the kublet config on azure and gcp

stonepreston:
When kubelet is running on control plane nodes, it needs to have a cloud config file written to its file system. [The worker charm](https://github.com/charmed-kubernetes/charm-kubernetes-worker/blob/4f3485f93152d6f836d06958cbea6a7ec7d250f2/reactive/kubernetes_worker.py#L1242-L1245) does this, but it is currently missing from the control plane charm. 

I tested this fix using GCP and can confirm it worked. I did not test on azure.


kwmonroe (approved):
LGTM; please link an LP bug for k-c-p 1.24 so this gets bubbled up into ck 1.24 release notes.

PR #224, opened 2022-05-05
Check for Ceph FSID before enabling Ceph CSI addons

Cynerva:


stonepreston (approved):
LGTM once CI goes green

addyess (approved):
double agree!

kwmonroe (approved):
I don't even care if CI goes green.  This looks amazing!

PR #225, opened 2022-05-09
Testing with python-3.10

addyess:


tox.ini
```diff
@@ -22,7 +22,8 @@ passenv = OS_* TEST_*
 deps =
     pyyaml
     pytest
-    charms.unit_test
+    #charms.unit_test
+    git+https://github.com/juju-solutions/charms.unit_test.git@feature/py310_support
```
addyess, line 26: Don't Merge with this change!!

addyess: Marking as draft so i remember not to merge with this change.  Feel free to review

kwmonroe: the [unit_test pr](https://github.com/juju-solutions/charms.unit_test/pull/19) is released as [v1.1.1](https://pypi.org/project/charms.unit-test/1.1.1/#files).  You can revert the `git+https` now if you'd like.

tests/unit/test_kubernetes_control_plane.py
```diff
@@ -322,15 +322,16 @@ def test_setup_auth_webhook_tokens(kcs, ctsar):
...
-@mock.patch.object(kubernetes_control_plane, "get_ingress_address", autospec=True)
-def test_ignore_vip(get_ingress_address, render):
-    get_ingress_address.return_value = "5.6.7.8"
+@mock.patch("pathlib.Path.exists", mock.Mock(return_value=False))
+@mock.patch("pathlib.Path.glob", mock.Mock(return_value=[]))
+@mock.patch("pathlib.Path.mkdir", mock.Mock(return_value=0))
+def test_ignore_vip():
+    kubernetes_control_plane.any_file_changed.return_value = False
+    mock_gia = kubernetes_control_plane.get_ingress_address
+    mock_gia.return_value = "5.6.7.8"
```
addyess, line 331: while testing with `python3.10` i noticed that this test wouldn't pass because mock.patch stated `kubernetes_control_plane.get_ingress_address` was already mocked.  Turns out that pre-python-3.10 this resulted in a mock of a mock -- so `get_ingress_address.assert_called_with` was a mock function and didn't actually test anything.  

Once unblocked by this double-mocking, i found the `register_auth_webhook(...)` hadn't been tested and needed to be updated to make it more testable.  This is the result of that work

kwmonroe (approved):

PR #229, opened 2022-06-23
Restart the auth-webhook service after certs change

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1978973 and https://bugs.launchpad.net/bugs/1956482

The auth-webhook service is configured to use a local server certificate [here](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/fb4460092b20e1151ee30672f3bdd3e4366717ed/templates/cdk.master.auth-webhook.service#L12-L13). If that cert changes, the auth-webhook service needs to be restarted, or it will continue to serve traffic using the old certificate.


addyess (approved):
🍰🚀👍🏻 LGTM 👍🏻🚀🍰

PR #230, opened 2022-06-24
share service_cidr over cni relation

addyess:
Using new relation endpoint method, share the `service-cidr` over the CNI relation

Depends on merging this first
* https://github.com/juju-solutions/interface-kubernetes-cni/pull/11/files

### DIFF
--- a/tests/unit/test_kubernetes_control_plane.py
+++ b/tests/unit/test_kubernetes_control_plane.py
@@ -1,3 +1,4 @@
+import contextlib
 import json
 from ipaddress import ip_interface
 from unittest import mock
@@ -41,6 +42,14 @@ def test_series_upgrade():
     assert kubernetes_control_plane.service_resume.call_count == 5
 
 
+@contextlib.contextmanager
+def endpoint_from_flag_reset():
+    restore = endpoint_from_flag.side_effect
+    endpoint_from_flag.mock_reset()
+    yield endpoint_from_flag
+    endpoint_from_flag.side_effect = restore
+
+
 @mock.patch("builtins.open", mock.mock_open())
 @mock.patch("os.makedirs", mock.Mock(return_value=0))
 def configure_apiserver(service_cidr_from_db, service_cidr_from_config):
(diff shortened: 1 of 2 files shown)


tests/unit/test_kubernetes_control_plane.py
```diff
@@ -76,31 +95,45 @@ def _svc(clusterIP):
...
-    configure_apiserver("10.152.183.0/24", "10.152.0.0/16")
+    cni = configure_apiserver("10.152.183.0/24", "10.152.0.0/16")
+    cni.set_service_cidr.assert_called_once_with("10.152.0.0/16")
     assert is_flag_set("kubernetes-control-plane.had-service-cidr-expanded")
     clear_flag("kubernetes-control-plane.had-service-cidr-expanded")
-    configure_apiserver("10.152.183.0/24,fe80::/120", "10.152.183.0/24,fe80::/112")
+    cni = configure_apiserver(
+        "10.152.183.0/24,fe80::/120", "10.152.183.0/24,fe80::/112"
+    )
+    cni.set_service_cidr.assert_called_once_with("10.152.183.0/24,fe80::/112")
```
addyess, line 136: @Cynerva 
I'm not sure if these two are sharing the correct `service-cidr`.  Can you confirm what the difference between the db value and config value and why one is picked over the other?

Cynerva: Most of the time, the two values are the same. However, when the charm starts kube-apiserver for the first time, the charm "freezes" the service CIDR (by setting the DB value) and no longer allows users to change it. The code around that is [here](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/fb4460092b20e1151ee30672f3bdd3e4366717ed/lib/charms/layer/kubernetes_control_plane.py#L405-L416).

We do this because changing the service CIDR in an existing cluster is extremely disruptive. It breaks services in Kubernetes. There is one exception: if the service CIDR change is an "expansion" (the new CIDR is a superset of the old CIDR) then the change is safe and the charm will allow it.

The assertions in these tests look correct to me. We're sharing the actual value that gets passed to kube-apiserver. That's what we need kube-ovn to match up with.

addyess: AH!!!  yes that explains what's going on.  Cool!  Thanks for the 👀

stonepreston (approved):
LGTM assuming integration passes

PR #231, opened 2022-06-30
Cni relation/image registry

stonepreston:
This PR sends the image-registry config over the CNI relation. Contingent on the merge of [this CNI interface PR](https://github.com/juju-solutions/interface-kubernetes-cni/pull/12).


reactive/kubernetes_control_plane.py
```diff
@@ -3704,6 +3704,19 @@ def request_ceph_permissions():
...
+def image_registry_changed():
+    registry = hookenv.config("image-registry")
+    # Share image-registry with cni requirers
+    cni = endpoint_from_flag("cni.available")
+    if cni:
+        endpoint_from_flag("cni.available").set_image_registry(registry)
+    else:
+        hookenv.log(
+            "CNI endpoint not available yet, waiting to set image registry data"
+        )
```
addyess, line 3717: so, if you configure this charm before the CNI interface is available, then the `image-registry` won't be set in the CNI relation.  

You should consider setting it during the `@when("cni.available")` hook as well

addyess (changes requested):
This is probably an issue

reactive/kubernetes_control_plane.py
stonepreston:
```
when_any("config.changed.image-registry", "cni.available")
```
is that not enough for that or would it need to be a separate hook? if the config changes before cni is available, it will log it, then when cni is available it should set the data

addyess: OH OH OH!  I missed the `when_any`

addyess (approved):
Alright, sorry, yeah looks good

PR #239, opened 2022-08-19
Fix for LP1986856

stonepreston:
There is probably a more reactive way of handling the 1.25 switch, but this seemed unobtrusive. Welcome to suggestions though.

Fixes https://launchpad.net/bugs/1986856


tests/integration/test_kubernetes_control_plane_integration.py
```diff
@@ -164,8 +165,13 @@ async def _auth_req(token, timeout=30):
     assert not any(await asyncio.gather(*tasks))
 
 
-async def test_pod_security_policy(ops_test, kubernetes):
+async def test_pod_security_policy(pytestconfig, ops_test, kubernetes):
     """Test the pod-security-policy config option"""
+    channel = pytestconfig.getoption("--snap-channel")
+    track, risk = channel.split("/", 1)
+    if float(track) >= 1.25:
```
Cynerva, line 172: If we cast versions to floats for comparisons, the comparisons will not always make sense. For example, `1.9` is greater than `1.25`, and `1.100` is less than `1.25`.

I think this particular case is unlikely to cause us problems for a long long time, but I think it's good practice to avoid casting versions to floats. I've had good results doing comparisons on tuples of integers instead:

```
if tuple(int(x) for x in track.split('.')) >= (1, 25):
```

tests/integration/conftest.py
```diff
@@ -36,7 +36,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--snap-channel",
         type=str,
-        default="1.24/stable",
+        default="1.25/candidate",
```
stonepreston, line 39: I am not sure there is a better way of doing this currently as the rc is only in this single channel. We would want to change this to latest/edge when 1.25 is put there I would think

kwmonroe, line 39:
```suggestion
```

kwmonroe (approved):

kwmonroe:
We'll follow up with this in a future PR to remove snap channel options (to ensure we use the default from k8s-core).

PR #240, opened 2022-08-22
Update series/snap channels to be pulled from bundle if not specified

stonepreston:
This PR updates the the snap channel and series command line arguments added previously to pull default values from the kubernetes core bundle. 

Unfortunately you cannot use fixtures in pytest hooks, so the default value of the command line argument was changed to an empty string, and a check was added to the series and snap channel fixtures to pull from the bundle if this option was empty.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -2,6 +2,7 @@ import logging
 import pytest
 import random
 import string
+import yaml
 
 from lightkube import KubeConfig, Client
 from lightkube.resources.core_v1 import Namespace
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -29,26 +30,48 @@ def pytest_addoption(parser):
...
+def k8s_core_bundle(ops_test):
+    return ops_test.Bundle("kubernetes-core", channel="edge")
+
+
+@pytest.fixture(scope="module")
+@pytest.mark.asyncio
+async def k8s_core_yaml(ops_test, k8s_core_bundle):
+    bundle_paths = await ops_test.async_render_bundles(k8s_core_bundle)
+    with open(bundle_paths[0], "r") as f:
+        return yaml.safe_load(f)
```
addyess, line 55:
```suggestion
async def k8s_core_yaml(ops_test, k8s_core_bundle):
    """Download and render the kubernetes-core bundle, return it's full yaml`
    bundle_path,  = await ops_test.async_render_bundles(k8s_core_bundle)
    return yaml.safe_load(bundle_path.read_text())
```

addyess, line 55:
```suggestion
async def k8s_core_yaml(ops_test, k8s_core_bundle):
    """Download and render the kubernetes-core bundle, return it's full yaml`
    bundle_path,  = await ops_test.async_render_bundles(k8s_core_bundle)
    return yaml.safe_load(bundle_path.read_text())
```

tests/integration/conftest.py
```diff
@@ -29,26 +30,48 @@ def pytest_addoption(parser):
...
-@pytest.fixture()
-def snap_channel(request):
-    return request.config.getoption("--snap-channel")
+@pytest.fixture(scope="module")
+def series(k8s_core_yaml, request):
+    series = request.config.getoption("--series")
+    if series:
+        return series
+    else:
+        return k8s_core_yaml["series"]
```
addyess, line 64: golf ⛳ 
```suggestion
     return series if series else k8s_core_yaml["series"]
```

tests/integration/conftest.py
```diff
@@ -29,26 +30,48 @@ def pytest_addoption(parser):
...
+
+
+@pytest.fixture(scope="module")
+def snap_channel(k8s_core_yaml, request):
+    channel = request.config.getoption("--snap-channel")
+    if channel:
+        return channel
+    else:
+        kcp = k8s_core_yaml["applications"]["kubernetes-control-plane"]
+        return kcp["options"]["channel"]
```
addyess, line 74: more golfing ⛳ 
```suggestion
    kcp = k8s_core_yaml["applications"]["kubernetes-control-plane"]
    return channel if channel else kcp["options"]["channel"]
```

addyess (approved):
Super nice.  i added some silly golfing exercises. 

I wonder if any of this is applicable for re-use across all our k8s charms.  

Maybe we should have a `charmed-kubernetes` pytest plugin for things like this

addyess (approved):
💯

PR #241, opened 2022-08-24
Fix version check when applying PSP

stonepreston:
The version check was previously failing with valid 1.24 versions like 1.24.4 since (1,24) <= (1, 24, x) returns False. The check was changed to check for versions less than (1, 25, 0) and unit tests were added for this case.


addyess (approved):
groovy.  i can dig it

addyess (approved):
Still dig it

PR #242, opened 2022-08-26
Append CSIMigrationvsPhere feature gate

stonepreston:
Adds `CSIMigrationvsPhere=false` to the feature gates for controller, scheduler, and api server. This allows the in-tree vsphere provider to create storage in 1.25+, but this flag is also desirable in lower versions as well.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -2459,6 +2459,7 @@ def configure_apiserver():
     ):
         api_opts["cloud-provider"] = "vsphere"
         api_opts["cloud-config"] = str(api_cloud_config_path)
+        feature_gates.append("CSIMigrationvSphere=false")
     elif is_state("endpoint.azure.ready"):
         api_opts["cloud-provider"] = "azure"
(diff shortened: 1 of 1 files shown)


reactive/kubernetes_control_plane.py
```diff
@@ -2672,6 +2674,11 @@ def configure_scheduler():
         feature_gates.append("CSIMigrationGCE=false")
     elif is_state("endpoint.azure.ready"):
         feature_gates.append("CSIMigrationAzureDisk=false")
+    elif is_state("endpoint.vsphere.ready") and get_version("kube-apiserver") >= (
+        1,
+        12,
+    ):
```
Cynerva, line 2680: `(1, 12)` - Is this meant to be `(1, 21)`?

I'd just skip this check myself, the 1.25 charms won't support k8s < 1.23 anyway.

Cynerva: Oh, I see, this is copied from the matching case in `configure_apiserver` and `configure_controller_manager` for consistency. Don't worry about it then, we can clean up these checks later.

Cynerva (approved):
LGTM, thanks

PR #244, opened 2022-08-30
Don't set CSIMigrationAWS or CSIMigrationGCE on k8s 1.25+

Cynerva:
https://bugs.launchpad.net/bugs/1988186


stonepreston (approved):
LGTM

addyess (approved):
I won't LGTM because it's sad to see this go.  But i will give my 👍🏻 as it is correct based on [1.25 changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.25.md#feature)

PR #245, opened 2022-09-06
LP#1988867 - properly parse the node status conditions

addyess:
parse node status from json when `NetworkUnavailable` does not exist in the status conditions


kwmonroe (approved):
Nice!

PR #248, opened 2022-09-14
include bootstrap option to use a pre-built focal image

addyess:
This change reduces the controller bootstrap time from 12min to 5min


mateoflorido (approved):
LGTM! 🎉

PR #249, opened 2022-09-22
Check Kube version before applying IPv6DualStack

jbrisko:
IPv6DualStack gate was removed in 1.25. I have added a version check not to inject that gate if the version is 1.25 or greater.


addyess:
Closing according to [this](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/250#issuecomment-1256505160)

Thanks for the contribution

PR #250, opened 2022-09-22
Remove IPv6DualStack feature gate

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1990455

The IPv6DualStack feature gate was removed in k8s 1.25 ([source](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.25.md#other-cleanup-or-flake)). Setting the feature gate now causes k8s services to fail.

This feature has been enabled by default since k8s 1.21 and GA since k8s 1.23 ([source](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)) so we shouldn't need to worry about backwards compatibility with this change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,3 +41,18 @@ jobs:
           bootstrap-options: "${{ secrets.FOCAL_BOOTSTRAP_OPTIONS }} --model-default datastore=vsanDatastore --model-default primary-network=VLAN_2763"
       - name: Run test
         run: tox -e integration
+      - name: Setup Debug Artifact Collection
+        if: ${{ failure() }}
+        run: mkdir tmp
+      - name: Collect Juju Status
+        if: ${{ failure() }}
+        run: |
+          juju status 2>&1 | tee tmp/juju-status.txt
+          juju-crashdump -s -m controller -a debug-layer -a config -o tmp/
+          mv juju-crashdump-* tmp/ | true
+      - name: Upload debug artifacts
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-run-artifacts
+          path: tmp
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -2327,8 +2327,6 @@ def configure_apiserver():
     # Handle static options for now
     api_opts["service-cluster-ip-range"] = service_cidr
     feature_gates = []
-    if kubernetes_common.is_dual_stack(cluster_cidr):
-        feature_gates.append("IPv6DualStack=true")
     api_opts["min-request-timeout"] = "300"
     api_opts["v"] = "4"
     api_opts["tls-cert-file"] = str(server_crt_path)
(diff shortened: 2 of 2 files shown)


addyess (approved):
LGTM too

addyess:
@jbrisko does this meet your needs?

kwmonroe (approved):
Thanks for the back-compat consideration (and upstream confirmation) that this is safe for n-2.

jbrisko:
@addyess LGTM. This should allow us to deploy 1.25 dual stack. Thanks for your help!

jbrisko:
How long does it take for this to make it to charmhub? Not sure what the release cycle is.

Cynerva:
We are actively working on a 1.25+ck2 release that contains this fix, which we expect to release by the end of next week (Oct 7). We're currently working through some issues with charm build failures due to recent changes to charm-tools, so it will probably be at least a few days more before we get this released.

In general, we don't have a consistent release cycle for bugfix releases - we release them "as needed" based on criticality of fixes to be released. We are discussing the possibility of a more regular release cadence but have not made a decision regarding that yet.

addyess:
@jbrisko i'm testing the candidate builds at the moment.  If you're DESPERATE the charm is available on `1.25/candidate` channel

jbrisko:
@addyess @Cynerva I have a deployment I am working on. I may pull candidate just to test that everything is working well. I can wait until Oct 7 to do a production deployment. Thanks for the info!

jbrisko:
@addyess @Cynerva  Looking at the Charmhub it seems that this hasn't been pushed in yet. Can you provide an update when to expect ck2? Thanks so much

addyess:
@jbrisko i don't understand?  

These changes are present in the 1.25/stable charm
```bash
juju download kubernetes-control-plane --channel=1.25/stable
mkdir /tmp/k8s-cp
unzip kubernetes-control-plane_*.charm -d /tmp/k8s-cp
grep IPv6DualStack /tmp/k8s-cp/reactive/kubernetes_control_plane.py | wc -l
```

Shows that `IPDualStack` is removed

jbrisko:
@addyess Thanks for the info. I was just watching https://charmhub.io/kubernetes-control-plane and it looked like nothing had been pushed in for awhile. I'll give it a shot. Thanks again!

jbrisko:
@addyess @Cynerva This seems to be working. However, I have a question, and maybe this isn't the best place to ask. If there is a better location, please point me in that direction. 

Essentially I'm trying to set up a 1.25 cluster with 2 kubeapi-load-balancers in front and keepalived handling the VIP between the two. I'm struggling to figure out which relations I need to set up there with the changes that have been made to the lb-consumers. Not sure I understand how each of those relations works. it seems like the load balancer is putting itself in as the endpoint in the Nginx config file. I'm sure this is just me not understanding the new relations.

Also, for some reason, when the charms generate the kubeconfig the server IP is wrapped in square brackets ( [ ] ). I think something isn't realizing that my API server uses an IPv4 address when using Dual Stack. I haven't dug into that issue much but figured you two might know precisely where to go for that bug.

Thanks!

jbrisko:
@Cynerva Thanks so much! This actually did get it to deploy and apply the virtual IPs correctly. Do you know if there are plans to update the keepalived charm, or is there a different way of doing basic HA that I should be looking at. Not sure I need a full corosync / pacemaker setup, but maybe that is the way.

Cynerva:
@jbrisko I'm not aware of any plans to update the keepalived charm beyond basic maintenance. I do want to at least update the documentation page for it though, and will try to do so in the next couple of days.

I think HAcluster is more widely used, if you want to give that a go. Docs are [here](https://ubuntu.com/kubernetes/docs/hacluster). It might be better maintained, but the whole STONITH thing has been a source of endless bugs that we still fight from time to time, so I'm a bit wary of it these days. Pick your poison, I guess. :P

We do consider both keepalived and hacluster to be fully supported. If you hit a critical issue with either, we will devote time to it.

PR #252, opened 2022-10-17
Addition of pip==20.0.2 and psutil==5.9.2

addyess:
In order to facilitate the charm installation of flit-core, pip version 20.0.2 is necessary See [LP#1991957](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1991957) for details

### DIFF
--- a/wheelhouse.txt
+++ b/wheelhouse.txt
@@ -3,6 +3,13 @@ gunicorn>=20.0.0,<21.0.0
 loadbalancer-interface
 typing_extensions<4.0
 
+# psutil 5.9.3 starts requirement of setuptools>=43
+# either we use this
+# psutil<5.9.3
+# or we use the default install in focal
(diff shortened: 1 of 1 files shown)


wheelhouse.txt
```diff
@@ -3,7 +3,18 @@ gunicorn>=20.0.0,<21.0.0
 loadbalancer-interface
 typing_extensions<4.0
 
+# psutil 5.9.3 starts requirement of setuptools>=43
+# either we use this
+# psutil<5.9.3
+# or we use the default install in focal
+# setuptools>=45.2.0
+setuptools>=45.2.0
```
addyess, line 11: We should really improve the comment here to explain WHY we're upgrading the setuptools.  I think it links nicely with the update of pip below as well.

stonepreston (approved):

PR #254, opened 2022-10-28
Don't reconfigure kube-apiserver when enable-metrics changes

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1995139

Reconfiguring kube-apiserver leads to disrupted access to the Kubernetes API. Don't reconfigure it, because nothing in the config actually changes.

This whole handler can be removed because `configure_cdk_addons` runs on every hook anyways.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -616,20 +616,6 @@ def send_cohorts():
     set_flag("kubernetes-control-plane.cohorts.sent")
 
 
-@when("etcd.available")
-@when("config.changed.enable-metrics")
-def enable_metric_changed():
-    """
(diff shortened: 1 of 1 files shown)


addyess:
@Cynerva are you saying [this handler](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/8164832cfeb4a054596cbf6ba1453b6a80e252cd/reactive/kubernetes_control_plane.py#L1674-L1680) will get run 

```python
@when(
    "kubernetes-control-plane.components.started",
    "leadership.is_leader",
    "leadership.set.cluster_tag",
)
@when_not("upgrade.series.in-progress")
def configure_cdk_addons():
```

PR #255, opened 2022-11-04
Remove `kube-dns` from `dns-providers`

mateoflorido:
## Purpose
Kube-DNS is included in K8s Legacy addons. We should remove it now that deployment on 1.25+ does not work.
## Changes
- Remove `Kube DNS` from the valid dns providers.
- Modify config.yaml to reflect the new providers list.
- Update README to reflect the changes.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -362,11 +362,7 @@ def check_for_upgrade_needed():
         leader_set(auto_storage_backend="etcd2")
 
     if is_leader and not leader_get("auto_dns_provider"):
-        was_kube_dns = hookenv.config().previous("enable-kube-dns")
-        if was_kube_dns is True:
-            leader_set(auto_dns_provider="kube-dns")
-        elif was_kube_dns is False:
-            leader_set(auto_dns_provider="none")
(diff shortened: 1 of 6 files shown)


reactive/kubernetes_control_plane.py
```diff
@@ -3366,8 +3356,6 @@ def get_dns_provider():
         if not dns_provider:
             if "core-dns" in valid_dns_providers:
                 dns_provider = "core-dns"
```
kwmonroe, line 3358: small knit; this `if` isn't needed anymore.
```suggestion
            dns_provider = "core-dns"
```

kwmonroe (approved):
LGTM with a tiny suggestion.

mateoflorido:
There are some issues with `setuptools` and `wheel` dependencies, affecting the build and run process of some charms in Focal (20.04). More information: [Chamhub](https://discourse.charmhub.io/t/cannot-install-dependencies-modulenotfounderror-no-module-named-setuptools-command-build/7374)

mateoflorido:
It looks like the dependency issue has been sorted out. Unpinning the dependency.

addyess (approved):
LGTM2

PR #257, opened 2022-11-21
presence of text in `stderr` in kubectl don't necessarily represent a failed command.

addyess:
While testing changed for containerd, it was discovered that 1.26/edge builds of kubectl presented with errors but completed successfully. There was `stderr` content while still fulfilling the requested command.  If `retcode == 0` but `stderr` had contents, the `auth-webhook` would fail without need.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -2587,7 +2587,9 @@ def configure_controller_manager():
         controller_opts["logtostderr"] = "true"
     if get_version("kube-controller-manager") >= (1, 26):
         controller_opts["requestheader-client-ca-file"] = str(ca_crt_path)
-        controller_opts["requestheader-allowed-names"] = "system:kube-controller-manager,client"
+        controller_opts[
+            "requestheader-allowed-names"
+        ] = "system:kube-controller-manager,client"
     controller_opts["kubeconfig"] = kubecontrollermanagerconfig_path
     controller_opts["authorization-kubeconfig"] = kubecontrollermanagerconfig_path
     controller_opts["authentication-kubeconfig"] = kubecontrollermanagerconfig_path
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -2585,11 +2585,6 @@ def configure_controller_manager():
     controller_opts["root-ca-file"] = str(ca_crt_path)
     if get_version("kube-controller-manager") < (1, 26, 0):
         controller_opts["logtostderr"] = "true"
-    if get_version("kube-controller-manager") >= (1, 26):
-        controller_opts["requestheader-client-ca-file"] = str(ca_crt_path)
-        controller_opts[
-            "requestheader-allowed-names"
-        ] = "system:kube-controller-manager,client"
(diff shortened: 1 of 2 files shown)


reactive/kubernetes_control_plane.py
```diff
@@ -2585,6 +2585,11 @@ def configure_controller_manager():
     controller_opts["root-ca-file"] = str(ca_crt_path)
     if get_version("kube-controller-manager") < (1, 26, 0):
         controller_opts["logtostderr"] = "true"
+    if get_version("kube-controller-manager") >= (1, 26):
+        controller_opts["requestheader-client-ca-file"] = str(ca_crt_path)
+        controller_opts[
+            "requestheader-allowed-names"
+        ] = "system:kube-controller-manager,client"
```
Cynerva, line 2592: Are these requestheader args needed now that the auth webhook is working? I'm struggling to fully understand the implications of this change and I'm not convinced that it's a change we want to make.

Cynerva: I'll try to educate myself on what changing this actually does. I've got a deployment coming up, will get back to you.

Cynerva: OK, yeah, I don't think we want to change this. Here's how it works:

The requestheader options are part of the [Kubernetes API aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/). We configure kube-apiserver with `--requestheader-client-ca-file` and `--requestheader-allowed-names`, to match the client certificate used in the `--proxy-client-key-file` and `--proxy-client-cert-file` options.

The requestheader options basically say: if I get a request, signed by the requestheader CA, whose common name is in the requestheader allowed names, then I will accept them as a valid proxy for the username/group specified the request has specified in the `X-Remote-User` and `X-Remote-Group` headers.

For the way we configure it, this means: I will accept proxied requests from kube-apiserver.

Now, that's not much use on kube-apiserver itself...it's not going to proxy to itself, is it? But what it does is create a ConfigMap named `extension-apiserver-authentication` that contains this info:

```
$ kubectl get configmap -n kube-system extension-apiserver-authentication -o yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: extension-apiserver-authentication
  namespace: kube-system
  ...
data:
  client-ca-file: |
    ...
  requestheader-allowed-names: '["system:kube-apiserver","client"]'
  requestheader-client-ca-file: |
    ...
  requestheader-extra-headers-prefix: '["X-Remote-Extra-"]'
  requestheader-group-headers: '["X-Remote-Group"]'
  requestheader-username-headers: '["X-Remote-User"]'
```

Then API extension services, such as metrics-server (and kube-controller-manager?) pull requestheader config from that configmap, so that *they* are now configured to accept proxied requests from kube-apiserver. Which allows them to act as extensions to the Kubernetes API.

So basically, we want kube-controller-manager requestheader config to match what's specified in kube-apiserver, and the easiest way to do that is to leave the configs blank and let it pull data from the ConfigMap.

addyess: awesome, i can confirm these changes aren't necessary for containerd integration tests as well.

Cynerva (approved):

PR #258, opened 2022-11-22
remove in-tree storage class definitions (AWS, GCE, and Azure)

addyess:
Starting in 1.25, the following in-tree storage classes could no longer be used, yet -- [cdk-addons still added them](https://github.com/charmed-kubernetes/cdk-addons/blob/9a72d030752d522e40f084ff0c9720b81a28be70/cdk-addons/apply#L187-L195) to the cluster.  

Let's just not enable them anymore


mateoflorido (approved):
LGTM!

PR #259, opened 2022-11-28
Fix Ceph pools flag

mateoflorido:
## Changes
We have been experiencing intermittent bugs in creating ceph pools. This is due to a mismatch in flag names, so this PR should address this issue.


addyess:
Yep, Looks like this was broken here:
* https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/208

addyess (approved):

PR #262, opened 2022-12-09
Update to select 1.26/stable snaps

addyess:
update defaults to 1.26/stable snaps and test against charms


kwmonroe (approved):
lgtm; these failures are expected as ck8s doesn't have a 1.26/stable channel let.

PR #264, opened 2022-12-13
remove CSIMigrationvSphere=True in 1.26.0 and beyond

addyess:
rollup all calls to `get_version(...)` to a single call per configure method
Use that `kube_version` variable to set/clear feature flags and arguments such as 
`CSIMigrationvSphere=false`


Cynerva (approved):

addyess:
the integration test failure here was after all the tests passed -- juju client complained a bit tearing down the machines

PR #265, opened 2023-02-06
Make CNI subordinate optional

Cynerva:
Depends on https://github.com/charmed-kubernetes/layer-kubernetes-common/pull/41

Make CNI subordinates optional for kubernetes-control-plane. CNI is still needed, but users will be able to deploy it themselves if desired.

If this charm is deployed without a CNI subordinate, then it will enter a Blocked status with the message: "Missing CNI relation or config". This message will go away after the user deploys their own CNI (i.e. `/etc/cni/net.d` becomes populated with CNI config).

Removed `bind-address` from kube-apiserver args. In my testing, whether it's unset, or set to `0.0.0.0`, or set to `::`, it always binds to *both* IPv4 and IPv6 addresses. So we can leave it unset.

Updated `send_data` handler to always get bind addresses for both IPv4 and IPv6.

Removed kube-controller-manager's `node-cidr-mask-size-ipv6` argument, for two reasons:
1. We don't configure kube-controller-manager with `allocate-node-cidrs=true`. Therefore, no node CIDRs are allocated to nodes, therefore this argument does nothing.
2. Even if we did allocate node CIDRs, this sets the prefixlen to match that of the whole cluster CIDR, which would cause the controller to allocate the entire cluster network to a single node. This is clearly not a desirable outcome for most clusters.


reactive/kubernetes_control_plane.py
```diff
@@ -237,6 +238,10 @@
     when_not="ceph-client.available",
     clear_flag="kubernetes-control-plane.apiserver.configured",
 )
+# when CNI becomes available, reconfigure k8s services with the cluster-cidr
+register_trigger(
+    when="cni.available", clear_flag="kubernetes-control-plane.components.started"
+)
```
kwmonroe, line 244: This tripped me up for a minute because we have a `when_not` trigger against `cni.available` above. Then i remembered/saw that triggers run once per endpoint, so these flags won't bounce around during every reactive run.

kwmonroe (approved):
LGTM

PR #266, opened 2023-02-08
override charmhelpers from layer-basic with charm-helpers default branch

addyess:
`kubernetes-control-plane` depends on `layer-vault-kv`.  It could be the only charm that uses this layer, i've not yet found another.

Before the 1.26 kubernetes release, we noticed that validation tests were failing when `hvac` (the vault client library) updated to `1.0.0` breaking compatibility with that layer 

We pinned hvac back to `< 1.0.0` in [`layer-vault-kv`](https://github.com/juju-solutions/layer-vault-kv/pull/16) and made a story to unpin ourselves later.

I have since updated [charmhelpers](https://github.com/juju/charm-helpers/pull/745), then made adjustments to [layer-vault-kv](https://github.com/juju-solutions/layer-vault-kv/pull/17) to also make use of newer `hvac`

This change alters the installed `charmhelpers` version in  `kubernetes-control-plane`  to prepare for future upgrades to layer-vault-kv by overriding `charmhelpers` present in layer-basic [`wheelhouse.txt`](https://github.com/juju-solutions/layer-basic/blob/master/wheelhouse.txt)

### DIFF
--- a/wheelhouse.txt
+++ b/wheelhouse.txt
@@ -2,6 +2,8 @@ aiohttp>=3.7.4,<3.8.0
 gunicorn>=20.0.0,<21.0.0
 loadbalancer-interface
 typing_extensions<4.0
+git+https://github.com/juju/charm-helpers.git#egg=charmhelpers
+
 
 # idna>=3.4 and beyond (needed by aiohttp and hvac)
 # requires flit-core for building its wheel.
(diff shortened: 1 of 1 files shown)


wheelhouse.txt
```diff
@@ -2,6 +2,8 @@ aiohttp>=3.7.4,<3.8.0
 gunicorn>=20.0.0,<21.0.0
 loadbalancer-interface
 typing_extensions<4.0
+git+https://github.com/juju/charm-helpers.git#egg=charmhelpers
+
```
kwmonroe, line 6:
```suggestion
```

kwmonroe (approved):
lgtm, suggested removing an extra blank line unless you put that there on purpose to slow the parser down or something.

PR #267, opened 2023-02-10
juju 3.0 renames the runner binary

addyess:
This service responds to a systemctl daemon which watches a set of files on the leader unit, and syncs them to other units in the cluster.

These changes ensure one of the binaries is appropriately called

Fixes: [LP:2006977](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2006977)

### DIFF
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,1 +1,4 @@
 docs/
+
+# wokeignore:rule=master
+templates/cdk.master.leader.file-watcher.*
--- a/templates/cdk.master.leader.file-watcher.service
+++ b/templates/cdk.master.leader.file-watcher.service
@@ -4,9 +4,8 @@ After=network.target
 
 [Service]
 Type=oneshot
-ExecStart=/bin/bash -c \
-    '/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh || \
-    /usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh'
+ExecStart=-/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=-/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
 
 [Install]
 WantedBy=multi-user.target
--- a/templates/cdk.master.leader.file-watcher.service
+++ b/templates/cdk.master.leader.file-watcher.service
@@ -4,7 +4,8 @@ After=network.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=-/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=-/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
 
 [Install]
 WantedBy=multi-user.target
--- a/templates/cdk.master.leader.file-watcher.service
+++ b/templates/cdk.master.leader.file-watcher.service
@@ -4,8 +4,9 @@ After=network.target
 
 [Service]
 Type=oneshot
-ExecStart=-/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
-ExecStart=-/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=/bin/bash -c \
+    '/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh || \
+    /usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh'
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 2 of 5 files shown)


.wokeignore
```diff
@@ -1 +1,4 @@
-docs/
\ No newline at end of file
+docs/
+
+# wokeignore:rule=master
+templates/cdk.master.leader.file-watcher.*
```
kwmonroe, line 4: TIL of this `.wokeignore` syntax.  neat!

templates/cdk.master.leader.file-watcher.service
```diff
@@ -4,7 +4,8 @@ After=network.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=-/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=-/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
```
kwmonroe, line 8: TIL of this `=-` syntax. neat!

kwmonroe (approved):

templates/cdk.master.leader.file-watcher.service
addyess: my only concern here is that `ExecStart=-/usr/bin/bad_command` will just log any errors but the service won't fail.  So, there will be no reporting of this service ACTUALLY failing if juju decides to rename a command again or alias the command.  

if they alias it (ie -- both binaries are present on the system) -- well then, BOTH commands would be run -- effectively calling this process twice. 

i also considered

```suggestion
ExecStart= \
  /usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh || 
  /usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
```

.wokeignore
addyess: this is actually the syntax to ignore a single baddy name... i'm actually woke-ignoring the `.wokeignore` file here.

templates/cdk.master.leader.file-watcher.service
addyess: hmm... something about the failed integration tests and these changes troubles me..

templates/cdk.master.leader.file-watcher.service
```diff
@@ -4,7 +4,9 @@ After=network.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh
+ExecStart=/bin/bash -c \
+    '/usr/bin/juju-exec {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh || \
+    /usr/bin/juju-run {{ unit }} /usr/local/sbin/cdk.master.leader.file-watcher.sh'
```
kwmonroe, line 9: @addyess how about this to address your concern with both `juju-[exec|run]` failing?  kinda ugly, but `bash -c` was the simplest way i found to chain ExecStarts.

Confirmed with edge charms + this PR:
```
ubuntu@juju-8ac7ee-6:~$ systemctl status cdk.master.leader.file-watcher.service 
○ cdk.master.leader.file-watcher.service - CDK control-plane leader file-watcher
     Loaded: loaded (/etc/systemd/system/cdk.master.leader.file-watcher.service; disabled; vendor preset: enabled)
     Active: inactive (dead) since Mon 2023-02-13 18:02:50 UTC; 17s ago
TriggeredBy: ● cdk.master.leader.file-watcher.path
    Process: 1482911 ExecStart=/bin/bash -c /usr/bin/juju-exec kubernetes-control-plane/1 /usr/local/sbin/cdk.master.leader.file-watcher.sh ||      /usr/bin/juju-run kubernetes-contr>
   Main PID: 1482911 (code=exited, status=0/SUCCESS)
        CPU: 78ms

Feb 13 18:02:50 juju-8ac7ee-6 systemd[1]: Starting CDK control-plane leader file-watcher...
Feb 13 18:02:50 juju-8ac7ee-6 bash[1482912]: /bin/bash: line 1: /usr/bin/juju-exec: No such file or directory
Feb 13 18:02:50 juju-8ac7ee-6 systemd[1]: cdk.master.leader.file-watcher.service: Deactivated successfully.
Feb 13 18:02:50 juju-8ac7ee-6 systemd[1]: Finished CDK control-plane leader file-watcher.
```

kwmonroe: Welp, this isn't holding up much better; looks like we're going for the 6hr timeout on tests.

addyess: this wasn't the reason behind the test-timeouts.  Turns out i had the wrong version of python libjuju selected in tox.ini.  I should have constrained it to `juju < 3.1`.  The 3.0 version of `juju` library has improved async handling on run events -- the very first `juju_run` call in the test was the point of the failure

templates/cdk.master.leader.file-watcher.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 
 # This script is invoked by cdk.master.leader.file-watcher.service
-
-if [ is-leader ]; then
+if [ $("is-leader") = "True" ];
```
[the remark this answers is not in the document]
addyess: sure enough
```bash
❯ juju run -u kubernetes-control-plane/1 -- 'is-leader; echo $?'
False
0

```

kwmonroe (approved):
re-approved again once more ftw

PR #270, opened 2023-02-16
sane perms on file-watcher; stop propagating unused CSVs

kwmonroe:
CK has used a webhook + secrets for authn since 1.19.  Stop sync'ing old file-based authn data around control-plane peers.  Also drop group/other write-privs from the file-watcher script.


addyess (approved):
🍰 🚀

PR #272, opened 2023-03-03
Remove "Waiting for workers." status

Cynerva:
This status made sense when no workers meant no Kubernetes nodes. Now that kubernetes-control-plane units are proper Kubernetes nodes, it no longer makes sense to have this status.

We are running into this with Cluster API testing when we have only a single control plane node and no workers.


Cynerva:
Fixes https://bugs.launchpad.net/bugs/2009190

stonepreston (approved):
LGTM

PR #276, opened 2023-03-14
LP#1989362: Catch `VaultNotReady` exceptions for logging and retry

addyess:
Quoting [LP#1989362](https://bugs.launchpad.net/charm-layer-vault-kv/+bug/1989362):

> It looks like both the `generate_encryption_key` handler[1] and the `_write_encryption_config` function[2] (or its caller) may need to be updated to catch `VaultNotReady`.

[1]: https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/7de2ed17e09c1709c49d72572eab299c5a64fcf0/reactive/kubernetes_control_plane.py#L3229
[2]: https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/7de2ed17e09c1709c49d72572eab299c5a64fcf0/reactive/kubernetes_control_plane.py#L3243

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -3211,15 +3211,19 @@ def create_secure_storage():
         # TODO: If Vault isn't available, it's probably still better to encrypt
         # anyway and store the key in plaintext and leadership than to just
         # give up on encryption entirely.
-        _write_encryption_config()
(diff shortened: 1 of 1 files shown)


Cynerva (approved):
Nice, LGTM :+1:

PR #279, opened 2023-04-04
reconfigure kubelet if our external cloud provider presence changes

kwmonroe:
We might configure the k-c-p kubelet before we know about an external cloud provider.  We should keep track of our xcp presence so we can reconfigure the kubelet when it changes.

### DIFF
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -3732,7 +3732,7 @@ def reconfigure_kubelet():
 @when("kubernetes-control-plane.kubelet.configured")
 def watch_xcp_for_changes():
     has_xcp = has_external_cloud_provider()
-    if data_changed("kubelet-has-xcp", has_xcp)
+    if data_changed("kubelet-has-xcp", has_xcp):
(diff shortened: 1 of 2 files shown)


reactive/kubernetes_control_plane.py
```diff
@@ -1526,8 +1526,11 @@ def send_api_urls():
     kube_control.set_api_endpoints(kubernetes_control_plane.get_api_urls(endpoints))
 
 
-def has_external_cloud_provider():
-    return bool(hookenv.relations().get("external-cloud-provider"))
+def has_external_cloud_provider() -> bool:
+    has_xcp = bool(hookenv.relations().get("external-cloud-provider"))
+    if data_changed("has-xcp", has_xcp):
+        set_flag("external-cloud-provider.changed")
```
kwmonroe, line 1532: On initial deploy, `data_changed` will be true, but this `e-c-p.changed` flag handler won't immediately reconfigure services since they won't be started yet.

On charm upgrade, the first call to this function will end up in this truthy `data_changed` and ultimately lead to a service reconfig, but that's ok because we want to ensure the `--cloud-provider=external` opt is set correctly on apiserver|cm|kubelet for all control-plane units going foward.

PR #280, opened 2023-04-04
add ignore pending pods config option

stonepreston:
This PR adds a config option that allows the user to specify pods to ignore during the check for pending kube-system pods. This is useful for CAPI as many pods are pending without a CNI deployed, which prevents the control plane from going active/idle. CoreDNS, for example, wont start without a CNI. 

The unit test for get_kube_system_pods_not_running was modified to test this new option. 

Additionally, the test_missing_cni was mistakenly placed under the TestSendClusterDNSDetail class on my previous PR. It was moved up with the rest of the tests, and mocks were added to account for the get_dns_provider and get_kube_system_pods_not_running calls since they contain some hookenv.config calls.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -367,5 +367,6 @@ options:
     type: string
     default: ""
     description: |
-      Space separated list of pod names that will be ignored when checking for running pods in kube-system. The pod names in kube system are compared
-      against this list, and if a kube-system pod name starts with one of the names provided, it will be ignored during the check.
+      Space separated list of pod names in the kube-system namespace to ignore
+      when checking for running pods. Any non-Running Pod whose name is on
+      this list, will be ignored during the check.
--- a/config.yaml
+++ b/config.yaml
@@ -363,3 +363,9 @@ options:
   labels:
     # Override default from layer-kubernetes-node-base config.yaml
     default: "node-role.kubernetes.io/control-plane="
+  ignore-kube-system-pods:
+    type: string
+    default: ""
+    description: |
+      Space separated list of pod names that will be ignored when checking for running pods in kube-system. The pod names in kube system are compared
+      against this list, and if a kube-system pod name starts with one of the names provided, it will be ignored during the check.
--- a/reactive/kubernetes_control_plane.py
+++ b/reactive/kubernetes_control_plane.py
@@ -243,11 +243,6 @@ register_trigger(
     when="cni.available", clear_flag="kubernetes-control-plane.components.started"
 )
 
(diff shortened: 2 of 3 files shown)


tests/unit/test_kubernetes_control_plane.py
```diff
@@ -525,6 +569,74 @@ def test_psp_config_1_25():
...
+@mock.patch(
+    "reactive.kubernetes_control_plane.get_kube_system_pods_not_running",
+    return_value=[],
+)
+@mock.patch("reactive.kubernetes_control_plane.get_dns_provider")
+@mock.patch(
+    "reactive.kubernetes_control_plane.hookenv.config",
+    side_effect=[False, True, "10.152.183.0/24", False],
+)
+def test_ignore_missing_cni(mock_config, mock_dns_provider, mock_get_pods):
```
stonepreston, line 582: this was in the wrong spot before, moved it up with the rest of the tests, and mocked out the dns call and the kube_system pod check since they contain hookenv.config calls and its easier if those are mocked out of the way

reactive/kubernetes_control_plane.py
```diff
@@ -2761,12 +2761,22 @@ def get_kube_system_pods_not_running():
...
     if result is None:
         raise FailedToGetPodStatus
 
+    # Remove pods whose names start with ones provided in the ignore list
+    pod_names_space_separated = hookenv.config("ignore-kube-system-pods")
+    ignore_list = pod_names_space_separated.strip().split()
+    result["items"] = [
+        pod
+        for pod in result["items"]
+        if not any(pod["metadata"]["name"].startswith(name) for name in ignore_list)
```
stonepreston, line 2777: using `startswith` to cover pod deployments that have the pod name with random chars on the end

config.yaml
```diff
@@ -363,3 +363,9 @@ options:
   labels:
     # Override default from layer-kubernetes-node-base config.yaml
     default: "node-role.kubernetes.io/control-plane="
+  ignore-kube-system-pods:
+    type: string
+    default: ""
+    description: |
+      Space separated list of pod names that will be ignored when checking for running pods in kube-system. The pod names in kube system are compared
+      against this list, and if a kube-system pod name starts with one of the names provided, it will be ignored during the check.
```
addyess, line 371:
```suggestion
      Space separated list of pod names in the kube-system namespace to ignore
      when checking for running pods. Any non-Running Pod whose name is on
      this list, will be ignored during the check.
```

PR #282, opened 2023-04-06
bump to cni 1.2.0 for k8s 1.27

kwmonroe:
LP: https://bugs.launchpad.net/charm-kubernetes-master/+bug/2007425

### DIFF
--- a/build-cni-resources.sh
+++ b/build-cni-resources.sh
@@ -29,8 +29,8 @@ mkdir "$temp_dir"
       -e GOOS=linux \
       -e GOARCH="$arch" \
       -v "$temp_dir"/cni-plugins:/cni \
-      golang:1.15 \
-      /bin/bash -c "cd /cni && ./build.sh && chown -R ${USER_ID}:${GROUP_ID} /cni"
+      golang:1.18 \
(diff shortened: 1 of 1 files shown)


mateoflorido (approved):
LGTM! The only doubt I had was about the use of go1.18. After seeing [this](https://github.com/containernetworking/plugins/blob/main/.github/workflows/test.yaml#L7), I now understand why.

Cynerva:
Hold please. I need to see if https://github.com/containernetworking/plugins/pull/633 affects this.

Cynerva:
Yeah, so, the flannel plugin was removed in [CNI plugins v1.0.1](https://github.com/containernetworking/plugins/releases/tag/v1.0.1). I'm pretty sure the flannel charm assumes that kubernetes-control-plane and kubernetes-worker will provide it. This will break our flannel support unless one of our charms (probably flannel) installs the [new standalone flannel CNI plugin](https://github.com/flannel-io/cni-plugin).

Cynerva (changes requested):
See above comment about flannel plugin.

kwmonroe:
Closing for now; will revisit with a flannel plan post-1.27.

PR #284, opened 2023-05-11
 [LP#1949913] Use a layer configurable option to change the secrets backend index

addyess:
Upgrades of the Kubernetes-Control-Plane charm allows for layer-vault-kv to disambiguate from other k8s-cp related to vault in a cross-model relation style

* set `secrets-backend-format` to use the model-uuid in the secrets backend
* testing upgrades

Associated with 
* https://github.com/openstack-charmers/charm-interface-vault-kv/pull/14
* https://github.com/juju-solutions/layer-vault-kv/pull/20

### DIFF
--- a/layer.yaml
+++ b/layer.yaml
@@ -44,6 +44,8 @@ options:
       - snap.kube-proxy.daemon
   hacluster:
     binding_address: 'kube-api-endpoint'
+  vault-kv:
+    secrets-backend-format: "charm-{model-uuid}-{app}"
 exclude:
   - .tox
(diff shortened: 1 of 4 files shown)


kwmonroe (approved):
Took me a while to grok how setting VaultAppKV[enc-key] was gonna keep k-c-p from re-running `generate_encryption_key()` when the _paths changed -- there's a ton of stuff going on between the setter (rehash + manage_flags) and all the atexits.  Really nice job keeping it clean.

PR #286, opened 2023-05-25
[LP#1949913] update delimieter to use a colon rather than a hyphen

addyess:
change the unit-name delimiter from - to :


addyess:
Closing because changing this from a hyphen to a colon results in an invalid URL path in vault charm

PR #287, opened 2023-06-12
Update renamed metrics to fix grafana dashboard. Fixes: LP#1956611

cjohnston1158:
[LP#1956611](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1956611) grafana dashboard metrics have either been renamed or gone missing since this template was created. 

This PR seems to fix the naming convention without any specific tests to confirm they won't change again in the future. 

[1] https://github.com/kubernetes/kubernetes/pull/71300/files
[2] https://github.com/kubernetes/kubernetes/pull/74418/files#diff-5a0d6b0e34ff7529a142eec3648065b523e9c69fba321eefd7eed5636b27ecd9L45-R45

### DIFF
--- a/templates/grafana/autoload/kubernetes.json
+++ b/templates/grafana/autoload/kubernetes.json
@@ -904,7 +904,7 @@
                         ],
                         "timeFrom":null,
                         "timeShift":null,
-                        "title":"kubernetes-master CPU idle",
+                        "title":"kubernetes-control-plane CPU idle",
                         "tooltip":{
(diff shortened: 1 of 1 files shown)


addyess (approved):
I never really look at these metrics, so there's no clear way to determine if these are correct now or in the future.

So -- it won't break any tests so LGTM

Looking for another +1 to merge

Cynerva (approved):
:+1: Thanks for the fix

PR #288, opened 2023-06-20
remove reactive flags checks that should no longer exist

addyess:
It's been 3 releases or more since these flags have been used or migrated, this is no longer necessary


Cynerva (approved):
LGTM

PR #289, opened 2023-06-20
Add cinder-availability-zone config

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1972861

This needs to be configurable in situations where Nova and Cinder have separate AZs.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -374,3 +374,9 @@ options:
       Space separated list of pod names in the kube-system namespace to ignore
       when checking for running pods. Any non-Running Pod whose name is on
       this list, will be ignored during the check.
+  cinder-availability-zone:
+    type: string
+    default: ""
+    description: |
+      Availability zone to use with Cinder CSI. This is passed through to the
+      parameters.availability field of the cdk-cinder StorageClass.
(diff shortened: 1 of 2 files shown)


addyess:
i'm not really thrilled with this.  It feels like exposing very openstack config settings through the control-plane config isn't quite right.  

But then again -- i'm not here offering a better solution:
maybe an addition to the `openstack` requires side relation so this can be set on the `openstack-integrator` charm?

Cynerva:
Yeah, existing precedent would have us make this a configuration option on openstack-integrator, which is sent to kubernetes-control-plane via the openstack-integration endpoint. A lot of existing options work this way.

I'm not a fan of the precedent because:
1. It adds more work for us (as you noted, requiring changes to openstack-integrator and interface-openstack-integration)
2. It makes the underlying architecture less obvious, and
3. It adds complexity to the charms that ultimately leads to more frequent bugs that are harder to root cause and fix (see for example [LP:1939596](https://bugs.launchpad.net/charm-openstack-integrator/+bug/1939596) and [LP:1965456](https://bugs.launchpad.net/charm-openstack-integrator/+bug/1965456))

I guess I'm just tired of making things complicated. I could be convinced otherwise if you think going through openstack-integrator is better.

> It feels like exposing very openstack config settings through the control-plane config isn't quite right.

I agree, but, I think it's a symptom of the fact that the architecture isn't quite right, and the solution is to fix the architecture. We're missing a production-ready cinder-csi charm. Long term, that's where this config needs to go.

addyess (approved):
I can't fault you for stating the obvious here that the config applies only to cdk-addons -- and since that's managed ultimately by the control plane charm -- the setting can go here.

PR #290, opened 2023-06-29
Generating openstack cloud config requires at least one relation endpoint

addyess:
When removing the `opentstack` endpoint from `kubernetes-control-plane`, the `relation-departed` hook fires before `relation-broken` hook.  

When running the `relation-departed` hook on the last endpoint -- the relation still exists, but there are no joined units.  It then becomes impossible to `generate_openstack_cloud_config` and the charm goes into `Error` state

Adding this simple check ensures there is at least one endpoint in the relation


mateoflorido (approved):
LGTM ❇️

PR #291, opened 2023-07-13
add cnis that require privileged containers to the auto list

addyess:
calico, cilium, and kube-ovn in 1.28 will require that the api server permit privileged containers

Let's add these to the auto list so we have seemless upgrades to 1.28 without requiring users to set the `allow-privileged` flag on the control plane charm

we can discover which cnis are connected using the cni-conf-file provided in the [cni relation](https://github.com/juju-solutions/interface-kubernetes-cni/blob/master/provides.py#L56-L76)


mateoflorido (approved):
LGTM!

PR #293, opened 2023-07-21
Handle certificates relation

Cynerva:
Depends on https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps/pull/2

This adds the necessary bits to relate to easyrsa/vault, request client and server certificates, and to write those certificates to disk.


kwmonroe (approved):
After a double-take to figure out where those config opts went prior to this PR, i was reminded we're in the `ops` target branch here :)

LGTM, parity++ ftw.

PR #299, opened 2023-08-25
Add container-runtime and cni endpoints

Cynerva:
Depends on:
- https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps/pull/5
- https://github.com/charmed-kubernetes/charm-lib-interface-container-runtime/pull/1
- https://github.com/charmed-kubernetes/charm-lib-interface-kubernetes-cni/pull/1

This adds the container-runtime and cni relation endpoints to the kubernetes-control-plane charm. We still can't run Calico, since we're not running kubelet yet, but we can at least test with Flannel.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -107,6 +107,13 @@ options:
     description: |
       Space-separated list of extra SAN entries to add to the x509 certificate
       created for the control plane nodes.
+  image-registry:
+    type: string
+    default: "rocks.canonical.com:443/cdk"
+    description: |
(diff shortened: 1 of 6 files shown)


Cynerva:
This is ready for review, but will need to have its requirements.txt updated after the dependency PRs are landed.

addyess (approved):
George this is looking so easy to read and follow. I need to go check out the new libraries. Love that they are packages

PR #300, opened 2023-09-01
Adds ipvs kernel modules to lxd profile

gustavosr98:
Required when using kube-proxy in ipvs mode
And the kubernenetes-control-plane charm is running on a LXD container

Fixes LP#2033682


Cynerva (approved):
LGTM, thanks!

PR #301, opened 2023-09-05
LXD profile for Calico required mount point /sys/fs/bpf

gustavosr98:
[LP#2034448](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2034448)

This is the workaround that worked for me
Not sure if this is the best approach to expose /sys/fs/bpf to the container of if it would be better to change the [mount point of the Calico pod](https://github.com/charmed-kubernetes/charm-calico/blob/main/upstream/calico/manifests/3.25.1/calico-etcd.yaml#L319)


mateoflorido (approved):
I think the `lxc.mount.auto=proc:rw sys:rw` configuration should also mount the `bpf` filesystem. However, +1. Thanks!

PR #308, opened 2023-10-31
Add hacluster integration

Cynerva:
Add hacluster integration to the ops version of kubernetes-control-plane. WIP.

This adds the `ha` relation endpoint and two config options: `ha-cluster-vip` and `ha-cluster-dns`. The implementation in this PR is able to register VIPs or DNS records with hacluster. If used, then those VIPs/hostnames will be used for Kubernetes API endpoints in kubeconfigs used by kubelet, kube-proxy, and end-users.

However, this work so far is missing any sort of failover mechanism. If kube-apiserver goes down on the unit that is holding the VIP, it will continue to hold that VIP. In the reactive charm, failover was handled by the charm during update-status hooks, where it would check the status of control-plane systemd services and update Pacemaker node status accordingly. See [here](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/9ca52889800937509bd0065d285c6646e04cb745/reactive/kubernetes_control_plane.py#L869-L892). This is obviously not ideal since it can mean failovers take up to 5 minutes with default Juju configuration, and could stop occurring entirely if the charm is in a bad state.

Prior to that, we used to register the systemd services to hacluster. However, that was removed because of a long history of bugs involving Pacemaker taking control of the systemd services and failing to run them when it should. See [here](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/190).

At this point, I see three potential ways to resolve this:
1. Handle failover during the update-status hook the same way the reactive charm does
2. Investigate hacluster/pacemaker configuration to see if there is a way to have it monitor the Kubernetes API without taking control of the systemd service
3. Deprecate hacluster support in kubernetes-control-plane entirely, and instead require kubeapi-load-balancer + hacluster to used for any solutions requiring HA

I suspect it may also be worth talking to the OpenStack team to figure out what the trajectory of the hacluster charm is. Will it be getting an ops uplift?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ charm-lib-node-base @ git+https://github.com/charmed-kubernetes/layer-kubernetes
 charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
 cosl == 0.0.7
 gunicorn >= 20.0.0,<21.0.0
-interface_hacluster @ git+https://github.com/openstack/charm-interface-hacluster
+interface_hacluster @ git+https://github.com/charmed-kubernetes/charm-interface-hacluster
 jinja2
 loadbalancer_interface
 ops >= 2.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ charm-lib-node-base @ git+https://github.com/charmed-kubernetes/layer-kubernetes
 charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
 cosl == 0.0.7
 gunicorn >= 20.0.0,<21.0.0
+interface_hacluster @ git+https://github.com/openstack/charm-interface-hacluster
 jinja2
 loadbalancer_interface
 ops >= 2.2.0
--- a/src/charm.py
+++ b/src/charm.py
@@ -29,10 +29,11 @@ from charms.kubernetes_libs.v0.etcd import EtcdReactiveRequires
 from charms.node_base import LabelMaker
 from charms.reconciler import Reconciler
 from cos_integration import COSIntegration
+from hacluster import HACluster
 from k8s_api_endpoints import K8sApiEndpoints
 from kubectl import kubectl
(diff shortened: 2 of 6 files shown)


Cynerva:
Fixes https://bugs.launchpad.net/bugs/2043695

I added basic VIP failover during the update-status hook, similar to how the reactive charm does it. This should be ready to go.

requirements.txt
```diff
@@ -10,6 +10,7 @@ charm-lib-node-base @ git+https://github.com/charmed-kubernetes/layer-kubernetes
 charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
 cosl == 0.0.7
 gunicorn >= 20.0.0,<21.0.0
+interface_hacluster @ git+https://github.com/openstack/charm-interface-hacluster
```
addyess, line 13: Let's use the fork since it's there.  We can keep it rev'd by branch

```suggestion
interface_hacluster @ git+https://github.com/charmed-kubernetes/charm-interface-hacluster
```

src/charm.py
```diff
@@ -443,6 +458,11 @@ def request_certificates(self):
 
         bind_addrs = kubernetes_snaps.get_bind_addresses()
         common_name = kubernetes_snaps.get_public_address()
+        config_addrs = [
+            address
+            for option in ["loadbalancer-ips", "ha-cluster-vip", "ha-cluster-dns"]
+            for address in self.config[option].split()
+        ]
```
addyess, line 465: is it reasonable to drop the address if they aren't configured?

```suggestion
            for address in self.config[option].split()
            if address
        ]
```

Cynerva: Sure. The `if address` filter will just ensure that `address` is never something falsey like `None` or `""`. I don't think the return value from `.split()` should ever include those, but, I do like defensive programming so I'll go ahead and include this.

addyess: too true.. it was a silly suggestion then :D

addyess (approved):
LGTM

PR #309, opened 2023-10-31
Add `cni-plugins` resource

mateoflorido:
## Summary
Add the CNI plugins for the Kubernetes Control Plane charm.

## Changes
- Add a script for building the resources.
- Update the metadata.yaml file.
- Handle the installation of the plugins in the host environment.


addyess (approved):
Yeah, i think this looks right

we'll need to adjust the charmcraft.yaml to build each arch into independent charms so the binary resources can be released independently with the charm

PR #310, opened 2023-11-02
Label control-plane nodes

addyess:
Use charm config to label the control-plane nodes with user supplied config

Automatically apply labels:
```yaml
  juju-application: <application-name>
  juju-charm: <charm-name>
  juju.io/cloud: <cloud-name if using external control-plane>
``` 

Depends on https://github.com/charmed-kubernetes/layer-kubernetes-node-base/pull/9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ __pycache__/
 *.py[cod]
 .idea
 .vscode/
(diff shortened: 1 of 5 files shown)


mateoflorido (approved):
LGTM

PR #313, opened 2023-11-10
Fix Metrics Endpoints

mateoflorido:


src/cos_integration.py
```diff
@@ -92,11 +117,32 @@ def get_metrics_endpoints(self) -> list:
...
+                f"kubelet-{path.split('/')[-1]}" if len(path.split("/")) > 2 else "kubelet",
+                path,
+                "https",
+                "localhost:10250",
+                [
+                    {"target_label": "metrics_path", "replacement": path},
+                    {"target_label": "job", "replacement": "kubelet"},
+                ],
+            )
             for path in kubelet_paths
```
addyess, line 130:
```suggestion
            JobConfig(
                f"kubelet-{metric}" if metric else "kubelet",
                path,
                "https",
                "localhost:10250",
                [
                    {"target_label": "metrics_path", "replacement": path},
                    {"target_label": "job", "replacement": "kubelet"},
                ],
            )
            for path in kubelet_metrics_paths
            if (metric := path.strip("/metrics")) is not None
```

requirements.txt
```diff
@@ -5,7 +5,7 @@ charm-lib-interface-external-cloud-provider @ git+https://github.com/charmed-kub
 charm-lib-interface-kube-dns @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kube-dns
 charm-lib-interface-kubernetes-cni @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kubernetes-cni
 charm-lib-interface-tokens @ git+https://github.com/charmed-kubernetes/charm-lib-interface-tokens
-charm-lib-kubernetes-snaps @ git+https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps
+charm-lib-kubernetes-snaps @ git+https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps@m/scheduler-auth
```
addyess, line 8: let's be careful not to merge yet with this branch in here

src/cos_integration.py
```diff
@@ -54,13 +57,10 @@ def _create_scrape_job(self, config: JobConfig, node_name: str, token: str) -> d
             "static_configs": [
                 {
                     "targets": [config.target],
-                    "labels": {"node": node_name},
+                    "labels": {"node": node_name, "cluster": self.charm.model.name},
```
addyess, line 60: it feels like this should be the [`cluster_name`](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/a41b8d7a7babace4f01d70910af3bfee92fc5465/src/charm.py#L346)

addyess (changes requested):
Maybe use the cluster_name rather than the model name

src/cos_integration.py
addyess: nice!

addyess (approved):
LGTM

PR #316, opened 2023-11-28
Fix AlertManager expr

mateoflorido:
### Summary
Fixes the `expr` for the alerting rules of the core K8s components.


Cynerva:
The expression change looks fine to me. But, it looks like this file was generated by [update_alert_rules.py](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/2632a45b36cb0c8b501353a076d5ab1c12e6c5fb/scripts/update_alert_rules.py) and if that script gets run again, this change will be lost.

IMO this has a high chance of being reverted by accident. I would recommend taking extra steps to protect this fix. Maybe save this change as a diff and have `update_alert_rules.py` apply the diff?

Cynerva (approved):
Nice, thanks :+1: LGTM

PR #318, opened 2023-12-01
[LP#2045432] pin yarl<1.9.3

addyess:
[LP#2045432](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2045432)

yarl 1.9.3 added `expandvars` in the `pyproject.toml ` but the charm tools aren't pulling this dep into the wheelhouse.

I'm guessing that outside of updating charm tools -- we should just pin yarl.


Cynerva (approved):
:+1:

PR #319, opened 2023-12-13
Pin config to 1.29/stable snaps and tests to 1.29/stable charms

addyess:
Per [release checklist](https://github.com/charmed-kubernetes/jenkins/blob/main/docs/releases/stable/index.md#pin-snap-channel-on-bundlescharms-in-the-release-branches)
* Expect integration tests to fail because 1.29/stable charms don't yet exist


mateoflorido (approved):
LGTM

PR #320, opened 2023-12-14
LP#2045009 Kubernetes apiserver certificate expiration is not monitored in 1.29 charms

mateoflorido:
## Summary
Add an alert rule for certificate expiration

### DIFF
--- a/src/prometheus_alert_rules/charmedKubernetes-prometheusRule.yaml
+++ b/src/prometheus_alert_rules/charmedKubernetes-prometheusRule.yaml
@@ -1,0 +1,11 @@
+groups:
+- name: ssl_cert_expiry_alert
+  rules:
+  - alert: SSLCertificateExpiringSoon
+    expr: probe_ssl_earliest_cert_expiry - time() < 86400 * 15
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      summary: "SSL certificate is expiring soon (instance {{ $labels.instance }})"
+      description: "SSL certificate for {{ $labels.instance }} expires in less than 15 days."


addyess (approved):
LGTM.  Thanks mateo!

nobuto-m:
Where do `probe_ssl_earliest_cert_expiry` metrics come from? Does Charmed K8s generate it out of the box?

yoshikado:
I have checked with the latest kubernetes-control-plane charm (rev344) which does have this fix, and deployed blackbox-exporter-k8s charm on microk8s where COS is deployed.

I can now confirm that the new Alert, SSLCertificateExpiringSoon, is added in Prometheus.
However, it seems that the expression in the alert doesn't match to any.

In my test environment, the SSLCertificateExpiringSoon alert looks for the following expression:

`probe_ssl_earliest_cert_expiry{juju_application="kubernetes-control-plane",juju_model="k8s",juju_model_uuid="fabfac13-87dc-44c8-88ed-14d90d72e9f5"} - time() < 86400 * 15`

However, the probe_ssl_earliest_cert_expiry metrics are showing the following results.

`probe_ssl_earliest_cert_expiry{instance="cos_91ea836d-2760-444d-88fa-3db484fa6b6c_blackbox-exporter", job="juju_cos_91ea836d_blackbox-exporter_prometheus_scrape_blackbox_k8s_endpoints", juju_application="blackbox-exporter", juju_charm="blackbox-exporter-k8s", juju_model="cos", juju_model_uuid="91ea836d-2760-444d-88fa-3db484fa6b6c", probe_target="https://172.16.20.30:6443"}`

It looks like it will only work when deploying blackbox-exporter-k8s on the same model where kubernetes is deployed.
However, as the blackbox-exporter charm is only for k8s, it can't be deployed on the same model as kubernetes.

screenshot-ssl-cert-alert
screenshot-ssl-cert-metric

PR #325, opened 2024-01-05
Adds kube-system unit status when the unit is stable

addyess:
Reinstates a check of the `kube-system` namespace to test if all the PODs within it are in a steady state.


kwmonroe (approved):
This is a nice reinstatement.  We've heard a few folks that like more control over when k-c-p reports Active vs Waiting based on kube-system workloads.

addyess:
@mateoflorido what do you think.  I didn't break any of the reconciliation protocols did i?

PR #327, opened 2024-01-17
Include lb-consumer relation's addresses when requesting certificates

addyess:
The `sans` argument of the certificate request lists which ips or hostnames are included in the certificate generated by the certificate authority (easyrsa or vault)

When we're using a loadbalancer for the k8s-api whether external or internal , the  address from the loadbalancer must be in the sans list.

* include those addresses in the `sans`
* if there are errors from the loadbalancer, report those with a blocking
* while waiting on the loadbalancer, reporting waiting status
* add other Maintenance status tasks

### DIFF
--- a/src/k8s_api_endpoints.py
+++ b/src/k8s_api_endpoints.py
@@ -31,12 +31,14 @@ class K8sApiEndpoints:
                     return build_url(addresses[0], 6443)
 
     def get_external_api_endpoint(self) -> Optional[str]:
+        """Endpoint address from the loadbalancer-external relation."""
         response = self.charm.lb_external.get_response("api-server-external")
         if not response or response.error:
             return None
         return response.address
 
     def get_internal_api_endpoint(self) -> Optional[str]:
+        """Endpoint address from the loadbalancer-external relation."""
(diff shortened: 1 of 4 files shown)


src/k8s_api_endpoints.py
```diff
@@ -30,6 +30,18 @@ def from_config(self) -> Optional[str]:
...
+        response = self.charm.lb_external.get_response("api-server-external")
+        if not response or response.error:
+            return []
+        return [response.address]
+
+    def get_internal_api_endpoints(self) -> List[str]:
+        response = self.charm.lb_external.get_response("api-server-internal")
+        if not response or response.error:
+            return []
+        return [response.address]
```
mateoflorido, line 43: Is it necessary for it to be a list containing only one element? This looks odd...

addyess:
I think the integration-tests are a bit flaky here.  There seems to be a race-condition where the kcp units go active-idle at the end of the reconcile loop -- but no update-status hook has been run

this means that the kube-system check is yet to run and it seems possible the node labels may not be applied. 

There may be more here that needs to be looked at.

mateoflorido (approved):
LGTM

PR #329, opened 2024-01-31
enforce juju >= 3.1

kwmonroe:
[LP:2051857](https://bugs.launchpad.net/charm-kubernetes-worker/+bug/2051857)

Ensure this charm only deploys in juju 3.1+ environments.

FYI to reviewers: series and assumes metadata keys are mutually exclusive. We can safely drop series here since that is now controlled by charmcraft.yaml's builds-on / runs-on directives.


addyess (approved):
LGTM!

PR #330, opened 2024-02-01
Handle unready relations with status.on_error

addyess:
Let's catch reconcile errors as they happen rather than let than the charm go into error

These changes aren't required to port back to release_1.29

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -487,7 +487,7 @@ class KubernetesControlPlaneCharm(ops.CharmBase):
     @status.on_error(ops.WaitingStatus("Waiting for certificate authority"))
     def request_certificates(self):
         """Request client and server certificates."""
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -80,24 +79,20 @@ def __init__(self, *args):
...
         client_cert = self.certificates.client_certs_map.get("system:kube-apiserver")
         server_cert = self.certificates.server_certs_map.get(common_name)
-
-        if not ca or not client_cert or not server_cert:
-            status.add(WaitingStatus("Waiting for certificates"))
-            return False
-
-        if not self.etcd.is_ready:
-            status.add(WaitingStatus("Waiting for etcd"))
-            return False
```
addyess, line 95: this is already asserted in a reconcile loop handler which executes before this one -- so it's not necessary

kwmonroe (approved):
Lgtm, thanks!

PR #332, opened 2024-02-06
Catch missing peer relation during charm upgrade

addyess:
Catches a missing peer relation during one of the upgrade-charm hook stages, and awaits its resolution. 

While preparing the `service-account-key` during a charm upgrade to be shared across the cluster via a peer relation, the standby units may come up without this relation and raise an exception while trying to access the `.data` attribute out of `None`

Similarly, the method `get_cluster_name()` could return `None` which failed on the non-leader units when trying to set the value in the `kube-control` relation.  It's possible to also trap those errors and prevent them from showing up as errored juju units.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,11 +17,15 @@ target-version = ["py38"]
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
+extend-exclude = ["__pycache__", "*.egg_info"]
+
+[tool.ruff.lint]
 select = ["E", "W", "F", "C", "N", "D", "I001"]
-extend-ignore = [
+ignore = [
     "D100",
     "D102",
     "D103",
+    "D107",
     "D203",
     "D204",
     "D213",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -33,12 +37,11 @@ extend-ignore = [
...
     "D409",
     "D413",
+    "E501",
 ]
-ignore = ["E501", "D107"]
-extend-exclude = ["__pycache__", "*.egg_info"]
 per-file-ignores = {"tests/*" = ["D100","D101","D102","D103","D104"]}
 
-[tool.ruff.mccabe]
+[tool.ruff.lint.mccabe]
```
addyess, line 44: All these changes are due to deprecation warnings in `ruff`.

They're drive by changes.

mateoflorido (approved):
Nice, I really like this pattern. LGTM!

addyess:
> Nice, I really like this pattern. LGTM!

@mateoflorido Then you'll like my next push

PR #334, opened 2024-02-13
open the apiserver port

kwmonroe:
[LP:2053048](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2053048)

Looks like we forgot to bring `open_port` over with the ops rewrite. You wouldn't notice it on a default `charmed-kubernetes` install because of k-l-b, but on `kubernetes-core`, you won't be able to access k-c-p outside of the cluster.

See bug for details.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -47,6 +47,8 @@ OBSERVABILITY_ROLE = "system:cos"
 class KubernetesControlPlaneCharm(ops.CharmBase):
     """Charmed Operator for Kubernetes Control Plane."""
 
+    APISERVER_PORT = 6443
+
     def __init__(self, *args):
         super().__init__(*args)
         self.cdk_addons = CdkAddons(self)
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -470,6 +476,14 @@ def reconcile(self, event):
...
             self.configure_observability()
             self.apply_node_labels()
+            self.open_ports()
+
+    def open_ports(self):
+        """Open control plane ports needed for remote access to the cluster."""
+        try:
+            self.unit.open_port("tcp", self.APISERVER_PORT)
+        except ops.ModelError:
+            log.warning(f"Failed to open port {self.APISERVER_PORT}. Will retry.")
```
addyess, line 486: the reconciler would rather you do:

```suggestion
    @status.on_error(ops.WaitingStatus("Waiting to open port"))
    def open_ports(self):
        """Open control plane ports needed for remote access to the cluster."""
        self.unit.open_port("tcp", self.APISERVER_PORT)
```

addyess: this decorator will catch that exception, log it, and hold off on "reconciling" the charm because of the failure.  This will mean it ACTUALLY is retried.   What's missing from this suggestion is your specific text around "Failed to open port" -- though a similar message and stack trace will be in your debug log and the charm will go into a `WaitingStatus`

addyess (changes requested):
One suggestion about how to use the reconciler.  thanks for this fix!

src/charm.py
kwmonroe: Splendid!  thanks for the suggestion.

addyess (approved):
LGTM

kwmonroe:
@addyess I figured if we're gonna open ports, we should probably close them too. My last commit includes close-port capabilities that keys off of reconcile's `api_dependencies_ready` check, but i dunno how useful that is.  Suggestions on a better place to use close-port are most welcome (observe the stop event maybe?).

FWIW, we didn't close-port in reactive, so if this is just wholly not worth it, i can back it out.

addyess:
@kwmonroe I saw the close-port merge as well.  I'm okay with it as well.  The reconciler should be observing the stop event so it should clean up properly

PR #335, opened 2024-02-16
[LP#2054126] Add get-kubeconfig action

addyess:
Simple, adds the get-kubeconfig action
[LP#2054126](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2054126)

PR #336, opened 2024-02-17
add choice betwee root and ubuntu config in kubectl helper

kwmonroe:
[LP#2054126](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2054126)

Follow on to #335. For the `get-kubeconfig` action, we need to use the public kubeconfig stored at `/home/ubuntu/config`. This has the external LB / apiserver endpoint configured.  The root kubeconfig uses localhost:6443, which is highly useless for admins that want to get at their cluster remotely :).

### DIFF
--- a/src/kubectl.py
+++ b/src/kubectl.py
@@ -23,7 +23,7 @@ def kubectl(*args, external=False):
     Returns stdout and throws an error if the command fails.
     """
     cfg = "/home/ubuntu/config" if external else "/root/.kube/config"
-    command = ["kubectl", f"--kubeconfig={cfg}"] + list(args)
+    command = ["kubectl", f"--kubeconfig={cfg}", *args]
     log.info("Executing {}".format(command))
     try:
         return check_output(command).decode("utf-8")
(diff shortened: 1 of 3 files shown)


src/kubectl.py
```diff
@@ -13,12 +13,17 @@ def get_service_ip(name, namespace):
...
 
+    By default, this function uses the root kubeconfig that points to the local apiserver.
+    Setting the 'external' parameter to 'True' will use the ubuntu config which points to
+    the external cluster endpoint.
+
     Returns stdout and throws an error if the command fails.
     """
-    command = ["kubectl", "--kubeconfig=/root/.kube/config"] + list(args)
+    cfg = "/home/ubuntu/config" if external else "/root/.kube/config"
+    command = ["kubectl", f"--kubeconfig={cfg}"] + list(args)
```
neoaggelos, line 26:
```suggestion
    command = ["kubectl", f"--kubeconfig={cfg}", *args]
```

neoaggelos (approved):

src/kubectl.py
kwmonroe: nice catch, thanks!

PR #343, opened 2024-04-03
[LP#2060070] Charm Actions Ops Port

addyess:
[LP#2060070](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2060070)

Port charm actions from reactive to ops that were identified as a gap during the upgrade. 

These are effectively straight port of admin actions available in the charm when it was reactive, addressing no issues with what was present in those previous actions like
* TTL for user token creation
* Validation of manifest application

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,4 +1,61 @@
-upgrade:
-    description: Upgrade the Kubernetes snaps
 get-kubeconfig:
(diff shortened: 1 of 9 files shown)


kwmonroe (approved):
wonderful, thank you!

PR #349, opened 2024-06-02
Untested port of cis-benchmark action to the kubernetes-control-plane

addyess:
### Overview
Addresses [LP#2044219](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2044219)

Re adds the cis-benchmark action from the reactive charm implementation

### Details
Takes a shot at porting the action from [layer-cis-benchmark](https://github.com/charmed-kubernetes/layer-cis-benchmark/blob/main/actions/cis-benchmark) into ops.  If it works out well -- likely convert to an ops version of this file in the original layer and included it as a pip package.

### DIFF
--- a/src/actions/csi_benchmark.py
+++ b/src/actions/csi_benchmark.py
@@ -1,0 +1,372 @@
+import contextlib
+import dataclasses
+import json
+import logging
+import os
+import shlex
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
+from typing import Optional
+
+import ops
+from charmhelpers.fetch.archiveurl import ArchiveUrlFetchHandler
+
+log = logging.getLogger(__name__)
+
+BENCH_HOME = Path("/home/ubuntu/kube-bench")
+BENCH_BIN = BENCH_HOME / "kube-bench"
+BENCH_CFG = BENCH_HOME / "cfg-ck"
+GO_PKG = "github.com/aquasecurity/kube-bench"
+RESULTS_DIR = "/home/ubuntu/kube-bench-results"
+
+# Remediation dicts associate a failing test with a Remedy to fix it.
+# Conservative fixes will probably leave the cluster in a good state.
+# Dangerous fixes will likely break the cluster.
+# Tuple examples:
+#  {'1.2.3': Remedy('manual -- we don't know how to auto fix this', None, None)}
+#  {'1.2.3': Remedy('cli', 'command to run', None)}
+#  {'1.2.3': Remedy('kv', 'snap', {cfg_key: value})}
+
+
+@dataclasses.dataclass
+class Remedy:
+    """Remedy class for benchmarking."""
+
+    type: str
+    command: Optional[str]
+    config: Optional[dict] = None
+
+    def run(self, test_num: int, test_remediation: str) -> int:
+        """Run the remedy command."""
+        if self.type == "manual":
+            log.info(
+                "Test %s: unable to auto-apply remedy.\n" "Manual steps:\n%s",
+                test_num,
+                test_remediation,
+            )
+        elif self.type == "cli":
+            cmd = shlex.split(self.command)
+            try:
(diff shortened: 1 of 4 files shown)


VariableDeclared:
any chance of movement on this?

src/actions/csi_benchmark.py
bschimke95, line 1: I guess the file name should be `csi_benchmark.py`, right?

src/actions/csi_benchmark.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def run(self, test_num: int, test_remediation: str) -> int:
+        """Run the remedy command."""
+        if self.type == "manual":
+            log.info(
+                "Test %s: unable to auto-apply remedy.\n" "Manual steps:\n%s",
+                test_num,
+                test_remediation,
+            )
+        elif self.type == "cli":
```
bschimke95, line 49: According to the comment above "kv" is also a possible type. Don't we need to handle this?

src/actions/csi_benchmark.py
```diff
@@ -0,0 +1,372 @@
...
+
+log = logging.getLogger(__name__)
+
+BENCH_HOME = Path("/home/ubuntu/kube-bench")
+BENCH_BIN = BENCH_HOME / "kube-bench"
+BENCH_CFG = BENCH_HOME / "cfg-ck"
+GO_PKG = "github.com/aquasecurity/kube-bench"
+RESULTS_DIR = "/home/ubuntu/kube-bench-results"
+
+# Remediation dicts associate a failing test with a Remedy to fix it.
```
bschimke95, line 24: nit: should this comment be part of the dataclass `Remedy`?

src/actions/csi_benchmark.py
```diff
@@ -0,0 +1,372 @@
...
+}
+ADMISSION_PLUGINS = {
+    "enable-admission-plugins": (
+        "PersistentVolumeLabel",
+        "PodSecurityPolicy," "AlwaysPullImages",
+        "NodeRestriction",
+    )
+}
+DANGEROUS = {
+    "0.0.0": Remedy("cli", 'echo "this is fine"', None),
```
bschimke95, line 74: Why do we have those dummy values in CONSERVATIVE and here?

src/actions/csi_benchmark.py
```diff
@@ -0,0 +1,372 @@
...
+        for name in files:
+            if name == filename:
+                log.info("Moving %s to %s", root, dest)
+                shutil.move(root, dest)
+                return
+    else:
+        raise ActionError(f"Could not find {filename} in {dirpath}")
+
+
+class CSIBenchmark(ops.Object):
```
bschimke95, line 135: nit: CSI and in comment it is CIS

bschimke95 (commented):
Some code comments. 
I cannot say much to the reactive (?) charm structure but I trust you that this is makes sense

src/actions/csi_benchmark.py
addyess: it doesn't look like the code path deals with that.  see https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/349/files#diff-28032d63f14eb463321b6f903bc8a8ed4f8ee3e76991509f56912ad7157efed7R238-R244

src/actions/csi_benchmark.py
addyess: I don't know -- this comes from the original code
https://github.com/charmed-kubernetes/layer-cis-benchmark/blob/main/actions/cis-benchmark#L45

addyess: these seemingly unnecessary values were originally introduced https://github.com/charmed-kubernetes/layer-cis-benchmark/pull/1

bschimke95 (approved):
LGTM

addyess:
@VariableDeclared should be in 1.31/edge tomorrow... i'll kick the build to be sure

PR #350, opened 2024-06-07
Partially support the keystone-credentials relation enough to uplift the keystone service

addyess:
## Overview
Allows admins to upgrade to the 1.29/stable relation without actually managing the keystone-credentials relation. 
Will detect any existing keystone auth service that were deployed via previous installations

Dependent on 
* https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps/pull/18
* https://github.com/charmed-kubernetes/kubernetes-docs/pull/845

### Changes:
* reintroduces the keystone-credentials interface
* recognize a keystone-auth-service installed via cdk-addons and render its webhook url into  the auth_webhook.py
* provides juju admin ability to set `authorization-webhook-config-file`
* adjust `kubectl()` helper method to retry only when the command fails (not when kubeconfig file is missing)
* adjust `kubectl()` helper method to reraise the CalledProcessError rather than a tenacity.RetryError
* block the charm when the keystone-credentials interface is related -- don't manage it any longer
* test...test...test... kubectl module

### DIFF
--- a/src/auth_webhook.py
+++ b/src/auth_webhook.py
@@ -36,11 +36,38 @@ class Secret:
     password: str
 
 
-def configure(
-    charm_dir, aws_iam_endpoint=None, custom_authn_endpoint=None, keystone_endpoint=None
-):
+def _uplift_keystone_endpoint() -> str:
+    """Uplift the keystone auth service from a cdk-addons installation."""
+    try:
+        keystone_auth_service = kubectl_get(
+            "service", "-n", "kube-system", "k8s-keystone-auth-service"
+        )
+    except CalledProcessError:
+        log.info("No k8s-keystone-auth-service to uplift")
+        return None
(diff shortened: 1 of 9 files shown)


src/auth_webhook.py
```diff
@@ -36,11 +38,38 @@ class Secret:
...
+        return None
+    cluster_ip, port = spec.get("clusterIP"), spec.get("ports")[0].get("port")
+    if not cluster_ip or not port:
+        log.error("No clusterIP or port found for k8s-keystone-auth-service")
+        return None
+    return f"https://{cluster_ip}:{port}/webhook"
+
+
+def _uplift_aws_iam_endpoint() -> str:
+    return None
```
addyess, line 65: Comment here with a `TODO` stating that there's future work present if necessary.

kwmonroe (approved):
lgtm

PR #353, opened 2024-06-17
Allow ceph-client to be related with a warning to no longer manage

addyess:
## Overview
Allow  upgrading from 1.28 charms and prior while related over the `ceph-client` relation.  Will warn that ceph-client is a deprecated relation

## Details
The kubernetes-control-plane charm should no longer manage ceph workloads, (ceph xfs, ceph rdb, ceph-fs) which instead should be delegated to the [ceph-csi](charmhub.io/cephi-csi) charm.  For those upgrading from 1.28 when this relation existed, the existing workloads they are running deployed via the older versions of the charm will remain and be unmanaged by the charm.  Post upgrade, one can remove the relation and the charm will not disturb its deployment.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -10,7 +10,7 @@ options:
       mode by default, unless certain circumstances are discovered
          * gpu hardware is detected on a worker node
          * openstack-integrator successfully related
-         * ceph-client successfully related
+
   api-extra-args:
     type: string
     default: ""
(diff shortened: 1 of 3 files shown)


kwmonroe (approved):
Very similar to the keystone-creds work; nice!  LGTM.

PR #354, opened 2024-06-20
Update charm to update the root kubeconfig with a current CA

addyess:
## Overview
Each config hook should do it's best to update the root kubeconfig to have a valid ca cert.

## Details
According to the bug [LP#2064305][], the charm code couldn't connect to the control-plane because 

```
2024-04-27 06:14:19 WARNING unit.kubernetes-control-plane/1.certificates-relation-changed logger.go:60 Unable to connect to the server: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This can be caused by the CA being invalid on the next charm hook.  This PR addresses the issue by ensuring the CA cert is correct on each possible reconcile loop

Draft until the following is merged
* https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps/pull/21

[LP#2064305]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/2064305

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ charm-lib-interface-external-cloud-provider @ git+https://github.com/charmed-kub
 charm-lib-interface-kube-dns @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kube-dns@main
 charm-lib-interface-kubernetes-cni @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kubernetes-cni@main
 charm-lib-interface-tokens @ git+https://github.com/charmed-kubernetes/charm-lib-interface-tokens@main
-charm-lib-kubernetes-snaps @ git+https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps@bug/lp2064305/support-updating-kubeconfig
+charm-lib-kubernetes-snaps @ git+https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps@main
 charm-lib-node-base @ git+https://github.com/charmed-kubernetes/layer-kubernetes-node-base@main#subdirectory=ops
 charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler@main
 interface_hacluster @ git+https://github.com/charmed-kubernetes/charm-interface-hacluster@main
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -4,7 +4,7 @@ charm-lib-interface-external-cloud-provider @ git+https://github.com/charmed-kub
 charm-lib-interface-kube-dns @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kube-dns@main
 charm-lib-interface-kubernetes-cni @ git+https://github.com/charmed-kubernetes/charm-lib-interface-kubernetes-cni@main
 charm-lib-interface-tokens @ git+https://github.com/charmed-kubernetes/charm-lib-interface-tokens@main
-charm-lib-kubernetes-snaps @ git+https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps@main
```
addyess, line 7: Must revert this line once merged

src/charm.py
```diff
@@ -330,26 +330,29 @@ def create_kubeconfigs(self):
...
+        else:
+            token = None
+
+        kubernetes_snaps.update_kubeconfig(
+            ROOT_KUBECONFIG,
+            ca=ca,
+            server=local_server,
+            token=token,
+            user="admin",
+        )
```
addyess, line 347: the crux of the fix is to ensure the ca is updated in the kubeconfig

src/kubectl.py
```diff
@@ -59,15 +60,18 @@ def kubectl(*args: str, external=False):
...
     try:
-        return check_output(command).decode("utf-8")
+        return check_output(command, stderr=PIPE).decode("utf-8")
     except CalledProcessError as e:
         log.error(
-            f"Command failed: {command}\nreturncode: {e.returncode}\nstdout: {e.output.decode()}"
+            f"Command failed: {command}\n"
+            f"returncode: {e.returncode}\n"
+            f"stdout: {e.stdout.decode()}\n"
+            f"stderr: {e.stderr.decode()}\n"
```
addyess, line 75: Extra debug on kubectl calls to print the stderr to the error log

src/charm.py
kwmonroe: ok - so this works because we may not have the api server up yet, hence can't verify `kubectl get secrets`, and if the CA changes during this time, our "bootstrap" client config wasn't getting updated because we were checking for "if file dne".

is that scenario what you're fixing with this?

kwmonroe (approved):
i added a comment to the crux of this fix just to make sure i understand why we need this.  assuming i got that scenario right, this LGTM.

and thanks for the extra kubectl debug!

src/charm.py
addyess: Right, according to the issue tracked in the bug, the certificate was invalid in the stack trace starting at 

```
auth_webhook.create_token(...)
   auth_webhook.get_token(...)
       auth_webhook.get_secrets(...)
            kubectl(...)
```

I surmised that the CA could be updated here to ensure we always have the right cert in place before checking the tokens.

PR #355, opened 2024-06-20
fix invalid dict merge in python3.8

addyess:
Cannot use the `|` to merge two dicts in python3.8 (focal).  Adjusting to a syntax which works in py3.8 and beyond


kwmonroe (approved):
Wow, nice find! +1 on this for py38 and beyond.

PR #358, opened 2024-06-25
Pin cdk-addons to the charm's channel and warn about cdk-addons deprecations

addyess:
## Addresses Issues
[LP#2070379](https://launchpad.net/bugs/2070379)

## Overview
Actively retire certain addons from cdk-addons by providing charm blocked state warnings for situations where users are using those features.

## Details
* pinning the cdk-addons snap based on the charm's release channel.  This will be determined from the charm branch the charm is built from (not the actual charmhub channel since that's not possible).  The risk (edge,candidate,beta,stable) will match the configured snap channel.  

example) So `juju config kcp channel=1.28/beta ` sets the snap channel.  If charm is built from the `release_1.30` branch, then cdk-addons `1.30/beta` will be installed along side the charm. 

* Readds the openstack relation is added back in purely so the user can upgrade and see a blocked message
* Readds the config option `enable-nvidia-plugin` whose only acceptable value is "false" in order to dismiss blocked message

Associated PRs:
* https://github.com/charmed-kubernetes/charm-lib-kubernetes-snaps/pull/24
* https://github.com/charmed-kubernetes/cdk-addons/pull/233


charmcraft.yaml
```diff
@@ -38,3 +38,10 @@ parts:
     build-packages: [git]
     prime:
       - templates/**
+  branch:
+    # save the branch name into the charm
+    after: [charm]
+    plugin: nil
+    source: ./
+    override-prime: |
+      git -C $CRAFT_PART_SRC branch --show-current >> $CRAFT_PRIME/templates/charm_branch
```
kwmonroe, line 47: clever!

kwmonroe (approved):
lgtm

PR #361, opened 2024-08-01
Improve error logging when failing to create a load-balancer

addyess:
### Overview
Addresses issue with [LP#2075524](https://bugs.launchpad.net/bugs/2075524)
Updated logging surrounding failed loadbalancer integration 

### Details
* Through the `loadbalancer_interface`, if a loadbalancer request fails via the integration, an error will be returned as an `error_message`.  Let's print that message rather than just that the charm is blocked.
* Improved the cloud integration logging to not mention an "unsupported cloud".  it's just a cloud that has no direct integrations necessary in the charm.


eaudetcobello (approved):

addyess:
@eaudetcobello thanks for catching up on the context in the LP and approving!

PR #362, opened 2024-08-16
Carefully use a status context on actions

addyess:
### Overview
Addresses [LP#2077189](https://bugs.launchpad.net/charm-kubernetes-worker/+bug/2077189)

Charm actions shouldn't use `status.context` because all that happens is that it ignores all exceptions and sets the unit status to Active/Idle

But the `upgrade` action has to use it since `kubernetes_snaps` depends on using the `status.add` method, so we have to carefully manage it

### Details
* A bunch of files are touched b/c `ruff` wants to test things harder.  
* The LP bug was discovered by @pedrofragola -- thanks so much
* But it turns out that using the `status.context` wipes out the charm's unit.status
   *  if `status.context` is used outside of a reconciled action, then we must run the reconciler again
   * MOST IMPORTANTLY -- we don't need WIPE out a WaitingStatus or BlockedStatus just because we ran an action. That's kinda bonkers

### DIFF
--- a/src/actions/restart.py
+++ b/src/actions/restart.py
@@ -1,14 +1,18 @@
-import charms.contextual_status as status
 import ops
 from charms import kubernetes_snaps
+
+SERVICES = {
+    "api-server": "snap.kube-apiserver.daemon",
+    "controller-manager": "snap.kube-controller-manager.daemon",
+    "kube-scheduler": "snap.kube-scheduler.daemon",
+}
 
 
 def restart_action(event: ops.ActionEvent):
     """Handle the upgrade action."""
-    with status.context(event.framework.model.unit):
-        kubernetes_snaps.service_restart("snap.kube-apiserver.daemon")
-        event.set_results({"api-server": {"status": "restarted"}})
-        kubernetes_snaps.service_restart("snap.kube-controller-manager.daemon")
-        event.set_results({"controller-manager": {"status": "restarted"}})
-        kubernetes_snaps.service_restart("snap.kube-scheduler.daemon")
-        event.set_results({"kube-scheduler": {"status": "restarted"}})
+    try:
+        for service, snap in SERVICES.items():
+            kubernetes_snaps.service_restart(snap)
+            event.set_results({service: {"status": "restarted"}})
+    except Exception as e:
+        event.fail(str(e))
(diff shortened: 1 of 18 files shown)


mateoflorido (approved):
LGTM!

src/actions/upgrade.py
```diff
@@ -3,8 +3,11 @@
...
+def upgrade_action(charm, event: ops.ActionEvent):
     """Handle the upgrade action."""
-    with status.context(event.framework.model.unit):
-        channel = event.framework.model.config.get("channel")
+    channel = event.framework.model.config.get("channel")
+    with status.context(charm.unit):
         kubernetes_snaps.upgrade_snaps(channel=channel, event=event, control_plane=True)
+    if isinstance(charm.unit.status, ops.ActiveStatus):
+        # After successful upgrade, reconcile the charm to ensure it is in the desired state
+        charm.reconciler.reconcile(event)
```
kwmonroe, line 13: ❤️

kwmonroe (approved):
Thanks @addyess. `ruff` bits were a bit rough to absorb in context, but it's an overall better charm for it. Invoking the reconciler after a successful active action is 💯 

I appreciate the deep dive into solving this upgrade ux; props to Pedro for helping get to the root cause!

src/actions/restart.py
```diff
@@ -1,14 +1,18 @@
...
-        kubernetes_snaps.service_restart("snap.kube-controller-manager.daemon")
-        event.set_results({"controller-manager": {"status": "restarted"}})
-        kubernetes_snaps.service_restart("snap.kube-scheduler.daemon")
-        event.set_results({"kube-scheduler": {"status": "restarted"}})
+    try:
+        for service, snap in SERVICES.items():
+            kubernetes_snaps.service_restart(snap)
+            event.set_results({service: {"status": "restarted"}})
+    except Exception as e:
+        event.fail(str(e))
```
addyess, line 18: @kwmonroe in case you missed it -- this runs this action in a try/catch rather than using `status.context` which would then require us to reconcile again.  I think this is a better approach to what the action intends

src/charm.py
```diff
@@ -132,6 +134,7 @@ def __init__(self, *args):
 
         # register charm actions
         actions = [
+            self.on.restart_action,
```
addyess, line 137: This actually REGISTERS the restart action -- it wasn't ever handled by the charm code !?!

PR #363, opened 2024-08-26
Mark the unit as waiting when kube-system pods aren't ready

addyess:
### Overview

Partially Addresses [LP#2032533](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2032533) by marking k-cp units in Waiting status when they have non-ready pods.  

### Details

* By checking the ready status of each of the containers in a pod, we can determine if the kube-system pods aren't ready and therefore mark the unit as "Waiting"
* This would have been helpful when spotting failing units caused by an incomplete cilium deployment


mateoflorido (approved):
LGTM! Thanks for implementing this :)

PR #366, opened 2024-12-05
[LP#2091126] Return the correct internal lb response

addyess:
### Overview
* Addresses [LP#2091126](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2091126)
* Reverts copy-paste mistake from #327

### DIFF
--- a/src/k8s_api_endpoints.py
+++ b/src/k8s_api_endpoints.py
@@ -38,8 +38,8 @@ class K8sApiEndpoints:
         return response.address
 
     def get_internal_api_endpoint(self) -> Optional[str]:
-        """Endpoint address from the loadbalancer-external relation."""
-        response = self.charm.lb_external.get_response("api-server-internal")
+        """Endpoint address from the loadbalancer-internal relation."""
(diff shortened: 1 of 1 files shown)


src/k8s_api_endpoints.py
```diff
@@ -38,8 +38,8 @@ def get_external_api_endpoint(self) -> Optional[str]:
         return response.address
 
     def get_internal_api_endpoint(self) -> Optional[str]:
-        """Endpoint address from the loadbalancer-external relation."""
-        response = self.charm.lb_external.get_response("api-server-internal")
+        """Endpoint address from the loadbalancer-internal relation."""
+        response = self.charm.lb_internal.get_response("api-server-internal")
```
addyess, line 42: INTERNAL loadbalancer IP should come from the INTERNAL relation, not the external relation

addyess:
I confirmed this uses AWS internal address for the kubeapi-load-balancer, not the external one.

mateoflorido (approved):

PR #369, opened 2025-01-21
Update Prometheus rule for cpu usage to apply to all nodes

drencrom:
Add specific labels allowing all juju applications to prevent COS from adding a specific `juju_application`. As this rule is added from the control plane, COS would filter only for the `kubernetes-control-plane` application and so the rule would not apply for nodes running on worker nodes.
See canonical/prometheus-k8s-operator#667

### DIFF
--- a/src/prometheus_alert_rules/kubernetesControlPlane-prometheusRule.yaml
+++ b/src/prometheus_alert_rules/kubernetesControlPlane-prometheusRule.yaml
@@ -1085,9 +1085,9 @@ groups:
 - name: k8s.rules
   rules:
   - expr: "sum by (cluster, namespace, pod, container) (\n  irate(container_cpu_usage_seconds_total{job=\"\
-      kubelet\", metrics_path=\"/metrics/cadvisor\", image!=\"\"}[5m])\n) * on (cluster,\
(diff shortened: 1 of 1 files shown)


mateoflorido:
Hello @drencrom!
Thank you for the contribution. This PR LGTM, but I wanted to point out that we pull the rules from the upstream repository using the Python scripts located in the `/scripts` directory of this repository. To manage custom patches for this rules we tipically use patches, similar to this one: https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/e0252948ab70479f241b666e01fd22f6b1a53b15/scripts/k8s-alert-rules.patch

Could you please create a patch and place it under the `/scripts` folder?

Thanks!

drencrom:
Adding some screenshots:

This is the list of pods generated from the control plane:
pod-list
As you see only the `openstack-cloud-controller-manager` is running in the control plane.

Here is a screenshot of the CPU usage dashboard for that pod:
pod-in-control-plane
As you see we have cpu usage information there even if it is zero.

Here is a screenshot of a pod not running on the control plane:
pod-outside-control-plane
As the pod does not have resource requests or limits the dashboard shows no data.

Here is another screenshot after applying this patch to the charm and generating some traffic to the pod:
pod-outside-control-plane-patched
You can see that now we can actually see cpu usage information.

mateoflorido (approved):
Thanks for the contribution, @drencrom!

PR #374, opened 2025-02-20
Do not mark succeeded or failed pods as unready

rapour:
Currently the `get_kube_system_pods_not_running` function checks whether the valid (i.e. pods in `Running`, `Succeeded`, or `Failed` phase) pods are ready and if not, returns them as `unready` for the charm to wait on them. However, completed pods are [marked as not ready](https://github.com/kubernetes/kubernetes/issues/52738). Therefore, the charm will wait for completed pods to become ready, which will never happen. The charm will recover after the `Completed` or `Failed` job's TTL passes and Kubernetes collects them, but the process will become lengthy and unpreadictable since it depends on jobs' TTL. 

Current PR suggests a fix to only check the `Running` pods for readiness.

### DIFF
--- a/src/k8s_kube_system.py
+++ b/src/k8s_kube_system.py
@@ -51,25 +51,26 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
         status = pod["status"]
         pod_phase, pod_reason = status["phase"], status.get("reason", "")
         if pod_phase == "Failed":
-            # Report failed pods as not running -- full stop
-            return True
-        if pod_phase == "Succeeded":
+            # Failed pods are not running -- full stop
+            not_running = True
+        elif pod_phase == "Succeeded":
             # Exclude Succeeded pods since they have run and done their work
-            return False
-        if pod_phase == "Running":
+            not_running = False
+        elif pod_phase == "Running":
             # Any Running phase pod with not ready containers, should be considered not running
             container_statuses = pod["status"].get("initContainerStatuses", [])
             container_statuses += pod["status"].get("containerStatuses", [])
-            return any(not status.get("ready", True) for status in container_statuses)
-        # Any other phase (Pending or Unknown) are not running if they aren't evicted
-        return pod_reason != "Evicted"
+            not_running = any(not status.get("ready", True) for status in container_statuses)
+        else:
+            # Any other phase (Pending or Unknown) are not running if they aren't evicted
+            not_running = pod_reason != "Evicted"
+        
+        if not_running:
+            pod_name, pod_ns = pod["metadata"]["namespace"], pod["metadata"]["name"]
+            log.warning("Pod/%s/%s in phase=%s is not running because of reason=%s", pod_ns, pod_name, pod_phase, pod_reason)
+        
+        return not_running
 
     not_running = [pod for pod in result["items"] if is_not_running(pod)]
-
-    log.info(
-        "Following pods are not running: {}".format(
-            ", ".join(pod["metadata"]["name"] for pod in not_running)
-        )
-    )
 
     return not_running
--- a/src/k8s_kube_system.py
+++ b/src/k8s_kube_system.py
@@ -47,26 +47,29 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
         )
     )
 
-    # Pods in phases such as ['Running', 'Succeeded', 'Failed']
-    # should not be considered as pending Pods.
-    valid_phases = ["Running", "Succeeded", "Failed"]
-
-    # Pods in ['Succeeded', 'Failed'] phases are not meant to become ready
-    def is_yet_to_be_ready(pod):
-
-        if pod["status"]["phase"] == "Running":
+    def is_not_running(pod) -> bool:
+        status = pod["status"]
+        pod_phase, pod_reason = status["phase"], status.get("reason", "")
+        if pod_phase == "Failed":
+            # Report failed pods as not running -- full stop
+            return True
+        if pod_phase == "Succeeded":
+            # Exclude Succeeded pods since they have run and done their work
+            return False
+        if pod_phase == "Running":
+            # Any Running phase pod with not ready containers, should be considered not running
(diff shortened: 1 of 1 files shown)


src/k8s_kube_system.py
```diff
@@ -51,17 +51,22 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
     # should not be considered as pending Pods.
     valid_phases = ["Running", "Succeeded", "Failed"]
 
+    # Pods in ['Succeeded', 'Failed'] phases are not meant to become ready
+    def is_yet_to_be_ready(pod):
+
+        if pod["status"]["phase"] == "Running":
+            container_statuses = pod["status"].get("initContainerStatuses", [])
+            container_statuses += pod["status"].get("containerStatuses", [])
+            return not(all(status.get("ready", True) for status in container_statuses))
```
addyess, line 60: maybe 
```suggestion
            return any(status.get("ready", False) for status in container_statuses)
```

src/k8s_kube_system.py
```diff
@@ -51,17 +51,22 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
     # should not be considered as pending Pods.
     valid_phases = ["Running", "Succeeded", "Failed"]
 
+    # Pods in ['Succeeded', 'Failed'] phases are not meant to become ready
+    def is_yet_to_be_ready(pod):
+
+        if pod["status"]["phase"] == "Running":
```
mateoflorido, line 57: We can use an early return, just to avoid the unnecessary indentation: 
```suggestion
    def is_yet_to_be_ready(pod):
        if pod["status"]["phase"] != "Running":
            return False
```

src/k8s_kube_system.py
```diff
@@ -51,17 +51,22 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
...
+            container_statuses = pod["status"].get("initContainerStatuses", [])
+            container_statuses += pod["status"].get("containerStatuses", [])
+            return not(all(status.get("ready", True) for status in container_statuses))
+
+        return False
```
addyess, line 62: Can we just combine `is_invalid` and `is_yet_to_be_ready` into one method for checking pods:

```python
def is_not_running(pod) -> bool:
   status = pod["status"]
   pod_status, pod_reason = status["phase"], status.get("reason", "")
   if pod_phase == "Failed":
     # Report failed pods as not running -- full stop
     return True
   if pod_phase == "Succeeded":
     # Exclude Succeeded pods since they have run and done their work
     return False
   if pod_phase == "Running":
      # Any Running phase pod with not ready containers, should be considered not running
      container_statuses = pod["status"].get("initContainerStatuses", [])
      container_statuses += pod["status"].get("containerStatuses", [])
      return any(status.get("ready", False) for status in container_statuses)
   # Any other phase (Pending or Unknown) are not running if they aren't evicted
   return pod_reason != "Evicted"
```


then below we filter with this method:

```python
    not_running = [pod for pod in result["items"] if is_not_running(pod)]
```

addyess (changes requested):
I think i'm suggesting something radical here... but maybe it makes it more readible and maintainable.  

I also wouldn't hate LOGGING which pods these are in each of the if blocks

src/k8s_kube_system.py
rapour: Thank you so much for the great suggestion, Done.

src/k8s_kube_system.py
```diff
@@ -47,21 +47,29 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
...
+        return pod_reason != "Evicted"
+
+    not_running = [pod for pod in result["items"] if is_not_running(pod)]
 
-    not_running = [pod for pod in result["items"] if is_invalid(pod) or not is_ready(pod)]
+    log.info(
+        "Following pods are not running: {}".format(
+            ", ".join(pod["metadata"]["name"] for pod in not_running)
+        )
+    )
```
addyess, line 73: Can we instead do our logging in the `is_not_running` method?

src/k8s_kube_system.py
```diff
@@ -47,21 +47,29 @@ def get_kube_system_pods_not_running(charm) -> Optional[List]:
...
+        if pod_phase == "Succeeded":
+            # Exclude Succeeded pods since they have run and done their work
+            return False
+        if pod_phase == "Running":
+            # Any Running phase pod with not ready containers, should be considered not running
+            container_statuses = pod["status"].get("initContainerStatuses", [])
+            container_statuses += pod["status"].get("containerStatuses", [])
+            return any(not status.get("ready", True) for status in container_statuses)
+        # Any other phase (Pending or Unknown) are not running if they aren't evicted
+        return pod_reason != "Evicted"
```
addyess, line 65: I'd really rather see the logging for each pod that's not running and why its not considered running.

```suggestion
        pod_phase, pod_reason = status["phase"], status.get("reason", "")
        if pod_phase == "Failed":
            # Failed pods are not running -- full stop
            not_running = True
        elif pod_phase == "Succeeded":
            # Exclude Succeeded pods since they have run and done their work
            not_running = False
        elif pod_phase == "Running":
            # Any Running phase pod with not ready containers, should be considered not running
            container_statuses = pod["status"].get("initContainerStatuses", [])
            container_statuses += pod["status"].get("containerStatuses", [])
            not_running = any(not status.get("ready", True) for status in container_statuses)
        else:
            # Any other phase (Pending or Unknown) are not running if they aren't evicted
            not_running = pod_reason != "Evicted"

        if not_running:
            pod_name, pod_ns = pod["metadata"]["namespace"], pod["metadata"]["name"]
            log.warning("Pod/%s/%s in phase=%s reason=%s", pod_ns, pod_name, pod_phase, pod_reason)
        return not_running
```

addyess (approved):
This is next review is a convenience ask.  I hope it makes sense

src/k8s_kube_system.py
rapour: Done

addyess (approved):
LGTM

PR #376, opened 2025-03-25
fix: scrape KSM only on the leader unit

mateoflorido:
## Overview
Scrape KSM metrics only on the control plane leader units.
## Rationale
Some reports mention that they see multiple times the same metrics but with different label, triggering a false positive alerts in AlertManager.
## Bug Reports
https://bugs.launchpad.net/charm-kubernetes-master/+bug/2104256

### DIFF
--- a/src/cos_integration.py
+++ b/src/cos_integration.py
@@ -139,7 +139,7 @@ class COSIntegration:
 
         kubelet_jobs = [
             JobConfig(
-                name=f"kubelet-{metric.replace('/', '-')}",
+                name=f"kubelet{metric.replace('/', '-')}",
                 metrics_path=metric,
                 scheme="https",
                 target="localhost:10250",
--- a/src/cos_integration.py
+++ b/src/cos_integration.py
@@ -100,45 +100,25 @@ class COSIntegration:
             "replacement": node_name,
         }
 
+        components = [
+            ("kube-proxy", "/metrics", "http", "localhost:10249"),
+            ("apiserver", "/metrics", "https", "localhost:6443"),
+            ("kube-scheduler", "/metrics", "https", "localhost:10259"),
+            ("kube-controller-manager", "/metrics", "https", "localhost:10257"),
+        ]
+
         kubernetes_jobs = [
             JobConfig(
-                "kube-proxy",
-                "/metrics",
-                "http",
-                "localhost:10249",
-                [{"target_label": "job", "replacement": "kube-proxy"}],
-            ),
-            JobConfig(
-                "apiserver",
-                "/metrics",
-                "https",
-                "localhost:6443",
-                [
-                    {
-                        "source_labels": ["job"],
-                        "target_label": "job",
-                        "replacement": "apiserver",
-                    },
-                    instance_relabel,
+                name=component,
+                metrics_path=path,
+                scheme=scheme,
+                target=target,
+                relabel_configs=[
+                    {"target_label": "job", "replacement": component},
+                    *([instance_relabel] if component != "kube-proxy" else []),
                 ],
-            ),
-            JobConfig(
-                "kube-scheduler",
-                "/metrics",
-                "https",
-                "localhost:10259",
-                [{"target_label": "job", "replacement": "kube-scheduler"}, instance_relabel],
-            ),
-            JobConfig(
-                "kube-controller-manager",
-                "/metrics",
-                "https",
-                "localhost:10257",
-                [
-                    {"target_label": "job", "replacement": "kube-controller-manager"},
-                    instance_relabel,
-                ],
-            ),
+            )
+            for component, path, scheme, target in components
         ]
         kubelet_metrics_paths = [
             "/metrics",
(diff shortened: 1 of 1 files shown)


src/cos_integration.py
```diff
@@ -100,45 +100,25 @@ def get_metrics_endpoints(self, node_name: str, token: str, cluster_name: str) -
             "replacement": node_name,
         }
 
+        components = [
+            ("kube-proxy", "/metrics", "http", "localhost:10249"),
+            ("apiserver", "/metrics", "https", "localhost:6443"),
+            ("kube-scheduler", "/metrics", "https", "localhost:10259"),
+            ("kube-controller-manager", "/metrics", "https", "localhost:10257"),
+        ]
```
addyess, line 108: Should we have this include another key like `instance-relabel : []` or instance-relabel :[instance_relabel]`?

src/cos_integration.py
```diff
@@ -100,45 +100,25 @@ def get_metrics_endpoints(self, node_name: str, token: str, cluster_name: str) -
...
-                [
-                    {"target_label": "job", "replacement": "kube-controller-manager"},
-                    instance_relabel,
+                name=component,
+                metrics_path=path,
+                scheme=scheme,
+                target=target,
+                relabel_configs=[
+                    {"target_label": "job", "replacement": component},
+                    *([instance_relabel] if component != "kube-proxy" else []),
```
addyess, line 118:
```suggestion
                    {"target_label": "job", "replacement": component}, *component["instance-relabel"]
```

src/cos_integration.py
addyess, line 108: doesn't this just beg for a type :D

```python
from dataclasses import dataclass

class Component(dataclass):
    name: str
    path: str
    schema: str
    host: str
    instance_relabel: List[???]
```

addyess (approved):
Cool,thanks @mateoflorido 

what do we need to do to test? 
Is there a LP for this change?
Do we need to backport to previous releases?

src/cos_integration.py
mateoflorido: Let's rollback to the JobConfig definition, adding a typing would introduce another dataclass similar to the JobConfig 😄

mateoflorido:
Verified that the metrics are indeed duplicated:
Screenshot From 2025-04-02 19-14-52

src/cos_integration.py
```diff
@@ -102,43 +102,33 @@ def get_metrics_endpoints(self, node_name: str, token: str, cluster_name: str) -
...
+                relabel_configs=[{"target_label": "job", "replacement": "apiserver"}, instance_relabel],
             ),
             JobConfig(
-                "apiserver",
-                "/metrics",
-                "https",
-                "localhost:6443",
-                [
-                    {
-                        "source_labels": ["job"],
```
addyess, line 118: did we lose `source_labels` for the `apiserver`?

src/cos_integration.py
```diff
@@ -149,39 +139,43 @@ def get_metrics_endpoints(self, node_name: str, token: str, cluster_name: str) -
...
-                path,
-                "https",
-                "localhost:10250",
-                [
-                    {"target_label": "metrics_path", "replacement": path},
+                name=f"kubelet-{metric.replace('/', '-')}",
+                metrics_path=metric,
+                scheme="https",
+                target="localhost:10250",
+                relabel_configs=[
```
addyess, line 146: ok, does this mean that that is no longer want `JobConfig` for just `kubelet` at the `/metrics` path?

Seems to me this would generate this in the list of `kubelet-Jobs`

```python
JobConfig(
    name="kubelet-",
    metrics_path=""
    ...
```

addyess (commented):
@mateoflorido I think i need help understanding these JobConfig objects a bit more.  I only spotted a few things that look like they may have significantly changed there

src/cos_integration.py
mateoflorido: Alright, so the name is not important here. So the charm is using the relabel_configs, so all scraped  metrics from Kubelet will have `job="kubelet"`, no matter the actual job name. This is the expected behavior, as the alerts target the `job="kubelet"` label.

addyess: neato

src/cos_integration.py
mateoflorido: This is not required. In the case of adding source_labels, Prometheus requires that a `job` label already exists in the original metrics to replace it. However, if we omit the `source_labels`, Prometheus will unconditionally set `job="apiserver"`, regardless of whether the label existed before.

addyess: it's wild how much this charm needs to know about prometheus in order to ship metrics to it.  But it seems important.  Thanks for the details.

mateoflorido:
After refreshing the snap with this patch, the metrics are correct without duplicates:
**Pod Count**
image
**Pod Count using Kubectl**
```
❯ kubectl get po -A --no-headers| wc -l
38
```
**Previous Query**
image

addyess (approved):
Still LGTM

PR #378, opened 2025-04-04
Add a script to invoke the config-changed event on changes to lo

addyess:
### Overview
observe routable changes to the `lo` interface and trigger juju config-changed events on the unit.


src/routabled.py
```diff
@@ -0,0 +1,67 @@
+import os
+from pathlib import Path
+from typing import Optional, TypedDict
+
+DEFAULT_TIMEOUT = 300  # 5 minutes
```
addyess, line 5: how long to wait running the config-changed hook before we decide it's not going to finish running the hook.  5m seems like plenty of time to run a charmed hook right?

addyess:
this turns out not to work as expected.  nothing i can do can get networkd-dispatcher to engage the `lo` interface

PR #380, opened 2025-04-10
bug: Validate allow-privileged configuration

addyess:
[LP#2058246](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2058246)
[LP#2106755](https://bugs.launchpad.net/charm-kubernetes-master/+bug/2106755)

### Overview
Validate charm configuration of `allow-privilged` to be either `false`, `true`, or `auto` and nothing else

### Details
* changing the default from `auto` to `false`  -- lets just go with secure by default shall we?
* allow for any-case of `true`/`false`
* stricter requirements on the value to be a string of `false`, `true`, or `auto`

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,15 +1,23 @@
 options:
   allow-privileged:
     type: string
-    default: "auto"
+    default: "false"
     description: |
       Allow kube-apiserver to run in privileged mode. Supported values are
-      "true", "false", and "auto". If "true", kube-apiserver will run in
-      privileged mode by default. If "false", kube-apiserver will never run in
-      privileged mode. If "auto", kube-apiserver will not run in privileged
-      mode by default, unless certain circumstances are discovered
-         * gpu hardware is detected on a worker node
-         * openstack-integrator successfully related
+      "true", "false", and "auto (deprecated)".
+
+      If "true", kube-apiserver will run in privileged mode.
+      If "false", kube-apiserver will never run in privileged mode.
+
+      Deprecated option:
+      If "auto", kube-apiserver will never run in privileged mode.
+      This mode was the previous default, so it's value is retained for backwards compatibility.
+
+      Prior to 1.29, kube-apiserver would run in privileged mode if the following conditions were met:
+      * gpu hardware is detected on a worker node
+      * openstack-integrator successfully related
+
+      Any other configuration will result in the charm blocked with an error message.
 
   api-extra-args:
     type: string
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -198,6 +198,13 @@ def service_extra_args(self, service_name, config_key) -> str:
         args = self.cis_benchmark.craft_extra_args(service_name, extra_args)
         return " ".join(f"{k}={v}" for k, v in args.items())
 
+    def allow_privileged(self) -> bool:
+        val = self.model.config["allow-privileged"].lower()
+        if val not in ["true", "false", "auto"]:
+            status.add(ops.BlockedStatus("Invalid config for allow-privileged"))
+            return False
+        return val == "true"
```
addyess, line 206: * Only `value == true` allows it
* invalid config sets unit to a blocked state, but the cluster can still go active with `allow-privileged=False`

src/charm.py
```diff
@@ -198,6 +198,13 @@ def service_extra_args(self, service_name, config_key) -> str:
         args = self.cis_benchmark.craft_extra_args(service_name, extra_args)
         return " ".join(f"{k}={v}" for k, v in args.items())
 
+    def allow_privileged(self) -> bool:
```
mateoflorido, line 201: This is only a suggestion, can we make this a `@property`? Or at least change the name to `get_allow_privileged`
```suggestion
    def allow_privileged(self) -> bool:
```

mateoflorido (approved):
LGTM, just a minor naming comments below.

src/charm.py
addyess: Done.  GREAT suggestion

config.yaml
```diff
@@ -1,15 +1,23 @@
 options:
   allow-privileged:
     type: string
-    default: "auto"
+    default: "false"
```
addyess, line 4: This may be a VERY bad default.  We need `true` here to run CNIs like `calico` and `cilium`.  Perhaps we should leave the default as `true` because otherwise so many things will suddenly start to fail

tests/data/charm.yaml
```diff
@@ -4,6 +4,8 @@ applications:
     charm: {{charm}}
     channel: null
     series: {{series}}
+    options:
+      allow-privileged: true
```
addyess, line 8: maybe we should just set the default to `true`??

mateoflorido (approved):
LGTM! ✨

PR #382, opened 2025-04-23
chore: 1.33 release readiness

HomayoonAlimohammadi:
### Overview
This PR updates the k8s version according to the [release doc](https://github.com/charmed-kubernetes/jenkins/blob/main/docs/releases/stable/index.md#pin-snap-channel-for-charms-in-the-release-branches)


HomayoonAlimohammadi:
Seems like the tests are failing because charms aren't available on 1.33/stable, which is true, but this step is supposed to be done before actually building 1.33 charms and populating the channels. Maybe we should merge this afterwards?

PR #383, opened 2025-04-23
Update cpu usage Prometheus rule (#369)

drencrom:
Add specific labels allowing all juju applications to prevent COS from adding a specific juju_application. As this rule is added from the control plane, COS would filter only for the kubernetes-control-plane application and so the rule would not apply for nodes running on worker nodes.


mateoflorido:
Closing in favor of manual cherry-pick

PR #385, opened 2025-04-30
Pin setuptools used for noble and jammy

addyess:
* pins setuptools used by python 3.10 and 3.12 
* overrides charmcraft 2.x's default of `setuptools-59.6.0`

Similar to:
* https://github.com/charmed-kubernetes/charm-kubernetes-worker/pull/191
* https://github.com/charmed-kubernetes/charm-kubernetes-e2e/pull/40
* https://github.com/charmed-kubernetes/charm-calico/pull/119
* https://github.com/charmed-kubernetes/charm-kubeapi-load-balancer/pull/46
* https://github.com/charmed-kubernetes/charm-kube-ovn/pull/60


charmcraft.yaml
```diff
@@ -534,12 +534,8 @@ bases:
 parts:
   charm:
     plugin: charm
-    build-packages:
-    - git
-    charm-python-packages:
-    - setuptools==79.0.0
-    prime:
-    - templates/**
```
mateoflorido, line 542: Are you sure we want to remove the `templates` from the priming step? Am I missing something?

addyess: from what i can tell, the charm plugin primed the templates folder.  But just for assurances i can leave it there

addyess: Here's a list of [CHARM_OPTIONAL_FILES](https://github.com/canonical/charmcraft/blob/2.7.5/charmcraft/const.py#L60) handled by charmcraft including `templates`.  I couldn't find specifically where they are included in the repo but i do this in the log when building:

```
'prime': ['templates/**', 'src', 'venv', 'dispatch', 'hooks', 'templates', 'README.md', 'lib', 'LICENSE', 'lxd-profile.yaml']},
```

indicating both the `templates/**` and `templates` folks are included.  Let's leave it in -- since it's not necessary to remove

PR #386, opened 2025-05-06
[release_1.32] Pin `setuptools` to `44.0.0`

mateoflorido:


addyess (approved):
Ya done good work here, Professor @mateoflorido

PR #387, opened 2025-05-06
[release_1.31] Pin `setuptools` to 44.0.0

mateoflorido:


addyess (approved):
🍄 "It’s-a perfect! Let’s-a merge!"

requirements.txt
```diff
@@ -16,11 +16,12 @@ ops.interface_azure @ git+https://github.com/charmed-kubernetes/interface-azure-
...
-loadbalancer_interface == 1.2.0
+loadbalancer_interface == 1.2.1
 ops == 2.12.0
 pydantic == 1.10.15
 tenacity == 8.2.3
 # kv and vault-kv
 hvac == 2.2.0
 # auth-webhook service
 gunicorn == 23.0.0
+setuptools == 44.0.0
```
addyess, line 27: Looks like we can drop this -- if we update charmcraft.yaml to have the `charm-binary-python-package` addition

mateoflorido: Sure, let's use the `charm-binary-python-packages` approach

PR #388, opened 2025-05-07
[release_1.32] Backport user create action ensures groups is a list of groups

mateoflorido:


pyproject.toml
```diff
@@ -9,15 +9,11 @@ show_missing = true
...
-# Formatting tools configuration
-[tool.black]
-line-length = 99
-target-version = ["py38"]
-
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
 extend-exclude = ["__pycache__", "*.egg_info"]
+target-version = "py38"
```
addyess, line 16: did we switch the 1.32 branch to ruff?

addyess (approved):
Other than the ruff changes, it looks good

pyproject.toml
addyess: [yes we did... nice](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/release_1.32/tox.ini#L27-L33)

PR #389, opened 2025-05-26
fix: support dual stack kubelet node ip addresses

rapour:
To addresss a similar issue as https://github.com/canonical/k8s-snap/issues/1448, this PR suggest changes to update the `--node-ip` flag of kubelet on control plane charm to include both ipv4 and ipv6 addresses, if present, which became a Cilium requirement with 1.17. It also checks that the IPs are not unspecified in case of a dual stack environment as per kubelet's requirement.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -784,14 +784,19 @@ class KubernetesControlPlaneCharm(ops.CharmBase):
                 line.split("=", 1) for line in output.strip().splitlines() if "=" in line
             )
 
-            if fields.get("ActiveState") == "failed" or fields.get("Result") == "exit-code":
+            active_state = fields.get("ActiveState")
+            result = fields.get("Result")
+            exec_main_status = fields.get("ExecMainStatus")
+            n_restarts = int(fields.get("NRestarts") or -1)
+
+            if active_state == "failed" or result == "exit-code":
                 return (
                     True,
-                    f"{service} has failed: ActiveState={fields.get('ActiveState')}, Result={fields.get('Result')}",
+                    f"{service} has failed: ActiveState={active_state}, Result={result}",
                 )
-            elif fields.get("ExecMainStatus") and fields["ExecMainStatus"] != "0":
-                return True, f"{service} Non-zero exit: ExecMainStatus={fields['ExecMainStatus']}"
-            elif int(fields.get("NRestarts", 0)) > 10:
+            elif exec_main_status and exec_main_status != "0":
+                return True, f"{service} Non-zero exit: ExecMainStatus={exec_main_status}"
+            elif n_restarts and n_restarts > 10:
                 return True, f"{service} is restarting repeatedly"
             return False, ""
         except subprocess.CalledProcessError as e:
--- a/src/charm.py
+++ b/src/charm.py
@@ -887,14 +887,18 @@ class KubernetesControlPlaneCharm(ops.CharmBase):
         uniq = {ipaddress.ip_address(addr) for addr in addresses}
         sorted_ips = sorted(uniq, key=lambda x: (x.version, x))
 
-        if len(sorted_ips) not in (1, 2) or (
-            len(sorted_ips) == 2 and (sorted_ips[0].version != sorted_ips[1].version)
-        ):
+        if not (0 < len(sorted_ips) <= 2):
             log.error(
-                "node-ips must contain either a single IP or a dual-stack pair of IPs (sorted_ips='%s')",
+                "node-ips must contain either a single IP or a dual-stack pair of IPs (ips='%s')",
                 sorted_ips,
             )
-            raise DualStackNodeIPError(f"{sorted_ips} is invalid")
+            raise DualStackNodeIPError(f"{sorted_ips} invalid length")
+        elif len(sorted_ips) == 2 and (sorted_ips[0].version != sorted_ips[1].version):
+            log.error(
+                "node-ips dual-stack pair of IPs must be different ip versions (ips='%s')",
+                sorted_ips,
+            )
+            raise DualStackNodeIPError(f"{sorted_ips} duplicate ip versions")
         # Dual-stack addresses (one IPv6 and one IPv4 address) must not contain unspecified IPs
         # https://github.com/kubernetes/kubernetes/blob/f6530285a85d6f4280711301613a7d3215a25818/staging/src/k8s.io/component-helpers/node/util/ips.go#L58
         elif len(sorted_ips) == 2 and any(addr.is_unspecified for addr in sorted_ips):
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,9 +66,7 @@ class MissingCNIError(Exception):
 
 
 class DualStackNodeIPError(Exception):
-    """Exception raised when there is an issue obtaining node IP(s) for kubelet
-    in a dual stack configuration.
-    """
+    """Exception raised when there is an issue obtaining node IP(s) for kubelet in a dual stack configuration."""
 
 
 def charm_track() -> str:
--- a/src/charm.py
+++ b/src/charm.py
@@ -64,10 +64,12 @@ class MissingCNIError(Exception):
 
     ERR = "Missing CNI relation or config"
 
+
 class DualStackNodeIPError(Exception):
-    """Exception raised when there is an issue obtaining node IP(s) for kubelet 
+    """Exception raised when there is an issue obtaining node IP(s) for kubelet
     in a dual stack configuration.
     """
+
 
 def charm_track() -> str:
     """Get the charm track based on the current charm branch.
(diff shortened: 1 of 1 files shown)


addyess (commented):
I have questions about this implementation

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
...
+
+        ipv4 = next((ip for ip in sorted_ips if ip.version == 4), None)
+        ipv6 = next((ip for ip in sorted_ips if ip.version == 6), None)
+
+        # Return a list with at most one IPv4 and one IPv6
+        node_ip_addresses = []
+        if ipv4:
+            node_ip_addresses.append(ipv4)
+        if ipv6:
+            node_ip_addresses.append(ipv6)
```
addyess, line 899: lets get rid of all of this

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
...
+
+        # Dual-stack addresses (one IPv6 and one IPv4 address) must not contain unspecified IPs
+        # https://github.com/kubernetes/kubernetes/blob/f6530285a85d6f4280711301613a7d3215a25818/staging/src/k8s.io/component-helpers/node/util/ips.go#L58
+        if len(node_ip_addresses) == 2 and any(addr.is_unspecified for addr in node_ip_addresses):
```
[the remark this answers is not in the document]
addyess: Lets use the same clever logic from kubernetes  `isv6(ip[0]) != ipv6(ip[1])`

```suggestion
        # Dual-stack addresses requires one IPv6 and one IPv4 address
        # https://github.com/kubernetes/kubernetes/blob/f6530285a85d6f4280711301613a7d3215a25818/staging/src/k8s.io/component-helpers/node/util/ips.go#L58
        if len(sorted_ips) == 2 and (sorted_ips[0].version == 6) != (sorted_ips[1].version == 6)
```

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
...
+            raise DualStackUnspecifiedIPError(f"{node_ip_addresses} contain unspecified IP")
+
```
addyess, line 905: maybe a separate check for `len(sorted_ips) > 2` or `len(sorted_ips) == 0`

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
...
+        node_ip_addresses = []
+        if ipv4:
+            node_ip_addresses.append(ipv4)
+        if ipv6:
+            node_ip_addresses.append(ipv6)
+
+        # Dual-stack addresses (one IPv6 and one IPv4 address) must not contain unspecified IPs
+        # https://github.com/kubernetes/kubernetes/blob/f6530285a85d6f4280711301613a7d3215a25818/staging/src/k8s.io/component-helpers/node/util/ips.go#L58
+        if len(node_ip_addresses) == 2 and any(addr.is_unspecified for addr in node_ip_addresses):
+            raise DualStackUnspecifiedIPError(f"{node_ip_addresses} contain unspecified IP")
```
addyess, line 904:
```suggestion
            raise DualStackUnspecifiedIPError(f"{sorted_ips} contain unspecified IP")
```

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
...
+
+        return [str(addr) for addr in node_ip_addresses]
```
addyess, line 906:
```suggestion
        return [str(addr) for addr in sorted_ips]
```

addyess (commented):
It's a bit of code golf here, but i like re-using the logic from kubernetes upstream to test for two different address versions

src/charm.py
```diff
@@ -877,6 +883,27 @@ def _check_kube_system(self):
                     msg = msg.format(len(unready), plural)
                     status.add(ops.WaitingStatus(msg))
 
+    def _get_node_ip_addresses(self) -> list[str]:
+        addresses = self.kube_control.ingress_addresses
+        uniq = {ipaddress.ip_address(addr) for addr in addresses}
```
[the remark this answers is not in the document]
rapour: The `kubelet` expects the IPs to not be unspecified only when `len(nodeIPs) == 2`. So I think this can create more restriction that is needed by kubelet. WDYT?

src/charm.py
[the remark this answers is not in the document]
rapour: Done, we have almost the same logic as the kubelet

src/charm.py
rapour: Done, thank you

src/charm.py
```diff
@@ -63,6 +64,10 @@ class MissingCNIError(Exception):
 
     ERR = "Missing CNI relation or config"
 
+class DualStackUnspecifiedIPError(Exception):
+    """Exception raised when there is an unspecified IP in a dual stack configuration."""
+
+    ERR = "Unspecified IP in dual stack configuration"
```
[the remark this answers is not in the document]
rapour: Done

src/charm.py
rapour: Error logs also added, done

src/charm.py
```diff
@@ -877,6 +882,30 @@ def _check_kube_system(self):
                     msg = msg.format(len(unready), plural)
                     status.add(ops.WaitingStatus(msg))
 
+    def _get_node_ip_addresses(self) -> list[str]:
```
mateoflorido, line 885: Wait... So this method raises an exception, and this one will go uncatched in the reconcile flow, (i.e., no `status.on_error` or similar construct in the `_configure_kubelet`. Should we add the decorator to that method? that way we can bubble up the charm status accordingly

addyess (approved):
Maybe just split up the invalid node-ip tests a bit more.  It's super defensive

src/charm.py
```diff
@@ -877,6 +882,30 @@ def _check_kube_system(self):
...
+        sorted_ips = sorted(uniq, key=lambda x: (x.version, x))
+
+        if len(sorted_ips) not in (1, 2) or (
+            len(sorted_ips) == 2 and (sorted_ips[0].version != sorted_ips[1].version)
+        ):
+            log.error(
+                "node-ips must contain either a single IP or a dual-stack pair of IPs (sorted_ips='%s')",
+                sorted_ips,
+            )
+            raise DualStackNodeIPError(f"{sorted_ips} is invalid")
```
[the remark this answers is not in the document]
rapour: Done

mateoflorido:
@rapour I was wondering if we could refine this a bit more.

First, I think we are offloading most of the checks already built into the core services, onto our charm. I strongly believe our charms should remain agnostic in that sense, since these changes may evolve in the future, and would require us to perform ongoing maintenance. Instead, we can just parse whatever is present in the environment and configure the kubelet with those arguments, even if they are not correct.

Second, to help catch those issues within the services, could you write a method that verifies the systemd services and surface any error it finds? For example, if the `node-ip` is misconfigured and this is detected by the logic embedded in the binary, the charm could notify the user that the core workloads are failing. This will be very useful and aligns well with the current spirit of the charming ecosystem.

So TL;DR:
- Remove the custom validation and just configure the service with the given values.
- Add to the `update_status` event a method to validate the status of the core services (e.g., `kubelet` or `apiserver`), and surface if any errors if finds through charm status.

src/charm.py
addyess: We should go the simplest route to collecting some addresses from the `kube-control` relation to use as args for node-ip, then let kublet fail if it doesn't like those arguments.

I think lines 885 -> 888 really accomplishes that.  

If kubelet fails to start, so be it -- we should have changes in `update_status` to determine if all the kubernetes services are actually running.  If they aren't running, our `update_status` should report that service is not yet running.  

The key problems with lines 889 -> 912 is that is duplicates logic that's really a part of `kubelet`.  If kubelet changes, this charm code also much change.  The charm should gracefully wrap `kubelet` -- not prevent it from ever failing to start by checking all the conditions it itself will check.  

I think we can safely say that if `kubelet` or any kubernetes service has failed to start once the charm is idle -- then `Blocked` is a likely diagnosis.  It's not likely time will solve the problem.  If ever hook and relation is done executing -- `Blocked` seems right.

@rapour Can you make the changes so it returns `sorted_ips` as is...

We can even delay the changes for the service checking to another PR.

addyess (changes requested):
I think simpler is a better approach here

src/charm.py
```diff
@@ -773,6 +774,37 @@ def request_certificates(self):
...
+                    True,
+                    f"{service} has failed: ActiveState={fields.get('ActiveState')}, Result={fields.get('Result')}",
+                )
+            elif fields.get("ExecMainStatus") and fields["ExecMainStatus"] != "0":
+                return True, f"{service} Non-zero exit: ExecMainStatus={fields['ExecMainStatus']}"
+            elif int(fields.get("NRestarts", 0)) > 10:
+                return True, f"{service} is restarting repeatedly"
+            return False, ""
+        except subprocess.CalledProcessError as e:
+            return True, f"Failed to check {service} status: {e.output.decode('utf-8')}"
```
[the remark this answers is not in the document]
rapour: Done

PR #391, opened 2025-05-29
chore: switch integration tests to noble

rapour:
Switches the integration test `series` from `jammy` to `noble`


addyess (approved):
I see the validation tests passed on ubuntu@24.04

PR #392, opened 2025-06-17
fix: set context for status in service check

rapour:
This PR suggests a fix by introducing status context when setting the status in service check


addyess (approved):
Thanks @rapour

PR #393, opened 2025-06-23
Use `node-base.address` library

addyess:
### Overview
* uses a library to get the `kube-control` address bindings to use with both the sans for this node's certificates as well as the node-ips used to register kubelet
* Uses the same library for api-server advertise_address

### Details
Deps:
* https://github.com/charmed-kubernetes/layer-kubernetes-node-base/pull/16

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -907,7 +907,7 @@ class KubernetesControlPlaneCharm(ops.CharmBase):
     def _check_kube_system(self):
         if not self.reconciler.stored.reconciled:
             # Bail, the unit isn't reconciled
-            log.info("Wait to check kube-system until reconciled")
+            log.info("Skipping kube-system check: unit is not yet reconciled.")
             return
 
         # only update the kube-system status under these conditions
--- a/src/charm.py
+++ b/src/charm.py
@@ -805,7 +805,7 @@ class KubernetesControlPlaneCharm(ops.CharmBase):
     def _check_core_services(self, services):
         if not self.reconciler.stored.reconciled:
             # Bail, the unit isn't reconciled
-            log.info("Wait to check core services until reconciled")
+            log.info("Skipping core services check: unit is not yet reconciled.")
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -804,6 +803,10 @@ def _service_has_failed(self, service):
             return True, f"Failed to check {service} status: {e.output.decode('utf-8')}"
 
     def _check_core_services(self, services):
+        if not self.reconciler.stored.reconciled:
+            # Bail, the unit isn't reconciled
+            log.info("Wait to check core services until reconciled")
+            return
```
addyess, line 809: only check the services if we are reconciled

src/charm.py
```diff
@@ -761,10 +760,10 @@ def request_certificates(self):
             "kubernetes.default",
             "kubernetes.default.svc",
             f"kubernetes.default.svc.{domain}",
+            *node_address.by_relation(self, "kube-control", True),
         ]
         sans += bind_addrs
         sans += config_addrs
-        sans += ingress_addrs
```
addyess, line 767: replace with the addresses from `node_address.by_relation(self, "kube-control", True)`

ktsakalozos (approved):
LGTM +1

src/charm.py
```diff
@@ -804,6 +803,10 @@ def _service_has_failed(self, service):
             return True, f"Failed to check {service} status: {e.output.decode('utf-8')}"
 
     def _check_core_services(self, services):
+        if not self.reconciler.stored.reconciled:
+            # Bail, the unit isn't reconciled
+            log.info("Wait to check core services until reconciled")
```
mateoflorido, line 808:
```suggestion
            log.info("Skipping core services check: unit is not yet reconciled.")
```

mateoflorido (approved):
LGTM. I left a nitpick to improve the language of the log message.

PR #396, opened 2025-07-23
[backport/release_1.33] Use `node-base.address` library

mateoflorido:
## Overview
Backport `node-base.address` to `release_1.33`


addyess (approved):
Nit about pinning to release versions of these libs...

